PR #13, opened 2015-01-06
Portuguese update

smarquespt:


luis-pereira:
@smarquespt Thanks. Landed at https://github.com/lxde/lximage-qt/commit/e007dc636683eef961e6c1a13fd42da9b9188075

PR #17, opened 2015-01-25
Corrected language code (de_DE -> de) of german translation, marked tran...

ThomasVie:
...slations as done.

### DIFF
--- a/src/translations/lximage-qt_de.ts
+++ b/src/translations/lximage-qt_de.ts
@@ -1,28 +1,28 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.0" language="de_DE">
+<TS version="2.1" language="de">
 <context>
     <name>LxImage::Application</name>
     <message>
-        <location filename="../application.cpp" line="112"/>
+        <location filename="../application.cpp" line="87"/>
         <source>Take a screenshot</source>
-        <translation type="unfinished">Erstelle ein Bildschirmfoto</translation>
+        <translation>Bildschirmfoto erstellen</translation>
     </message>
     <message>
-        <location filename="../application.cpp" line="113"/>
+        <location filename="../application.cpp" line="91"/>
         <source>[FILE1, FILE2,...]</source>
-        <translation type="unfinished">[Datei1, Datei2,...]</translation>
+        <translation>[Datei1, Datei2,...]</translation>
     </message>
 </context>
 <context>
     <name>LxImage::MainWindow</name>
     <message>
-        <location filename="../mainwindow.cpp" line="144"/>
+        <location filename="../mainwindow.cpp" line="147"/>
         <source>About</source>
-        <translation type="unfinished">Über</translation>
+        <translation>Über</translation>
     </message>
     <message>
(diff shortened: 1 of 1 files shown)


jleclanche:
I'm not sure the language code is incorrect here. I don't know how qt handles it though.

ThomasVie:
Now the language code corresponds to the file name.
Qt should create a _de.??? file. This will be taken if LANG is de or one of the more special ones (e.g. de_AT, de_CH, de_DE, ...) and a more special translation is not available.

ThomasVie:
LC_ALL is the correct environment variable.
Maybe you like to test the result with the following lines:
> env LC_ALL=ja lximage-qt
> env LC_ALL=de lximage-qt
> env LC_ALL=de_DE.UTF-8 lximage-qt
... insert the language code you like
instead of "env" you could use export, too.

pmattern:
Changes should be made to keep strings neutral with regards to "Du" vs. "Sie" ("pronominale Anredeform" vs. "Höflichkeitsform").
Affected lines and proposals: 9 → "Bildschirmphoto erstellen", 44 → "Datei öffnen"(should be adjusted in line 98 as well to keep the translation of "Open File" consistent), 55 → "Datei speichern" and 391 → "Bildschirmfoto erstellen".

"Preferences" in line 353 should be translated as "Einstellungen" as this is common practice in lots of other applications as well as in LXQt Application menu.

The multiple comments "Shortcuts werden automatisch übersetzt" should be in English. The meaning of the one in line 314 ("&apos;file&apos;") isn't obvious to me.

Regarding the language code German was used by all LXQt components I've tested so far when "de" was stated in *.ts files and LXQt was launched with LANG set to any of de_{AT,CH,DE}-UTF-8.
So using a language-specific code and omitting any country-specific one seems to work as expected.

paulolieuthier:
@pmattern do you think it's ready now?

pmattern:
I wasn't aware of that redundancy of term "File" as you state in lines 98 and 302, but you're absolutely right.
So it's okay in this translated file but indeed should be changed in lximage-qt_template.ts.

Regarding "Shortcuts werden automatisch übersetzt" you translated once but simply skipped all other appearances.
Not sure about the impact myself, but I guess you introduced that phrase for a reason...

Other than that GTM.

@paulolieuthier 
Would you mind me merging this? Just to go through the exercise as discussed yesterday. Maybe won't be today any more but won't forget it.

paulolieuthier:
@pmattern of course. I'll be glad to help you.

pmattern:
@ThomasVie

Sorry for the delay!

Are you sure about keyboard shortcuts getting translated automatically? I'm not an expert regarding this and just asking as I recently stumbled upon some strings like 'Ctrl' or 'Home' in PCManFM-Qt which had not been translated automatically.
Maybe it wouldn't hurt to quickly translate manually here as well?

While at it you may want to squash the commits into one to help us keep the commit history clean.

ThomasVie:
Squashed commits. Here is a screenshot showing the translated keyboard shortcuts,
so here it is no problem:

pmattern:
Translation of keyboard shortcuts seems indeed to get handled differently than in PCManFM-Qt.

Thanks for contributing and your patience.

Issue #22, opened 2015-02-16
Window size is not remembered


goetzc:
After reopening the program, the window size is not remembered, nor if it is maximized or restored. Tested using the latest commit 62ce73d.

jleclanche:
Fixed in #24

ralf3u:
choose an image (a screenshot in full-screen) in the file-manager
press enter
maximize the window
close the window
choose the same image in the file-manager
press enter
the window is maximized, so it works fine

now the issue:
choose an same image in the file-manager
press enter
the window is already maximized
press F11 to go in full-screen
press again F11 to leave full-screen
close the window
choose the same image in the file-manager
press enter
the window is not maximized

my 0.16.0 is old, but I still don’t know how to install the latest version;
if the issue does not appear in 0.17.0 any more, then I’m sorry for this comment

ralf3u:
Just for documentation: The content of the last comment was used for the issue https://github.com/lxqt/lximage-qt/issues/587.

PR #24, opened 2015-02-27
save and restore window size and maximized state

Caig:
To address issue #22, no gui option, more or less just copied from pcmanfm-qt.


jleclanche:
Thanks a lot!

PR #28, opened 2015-06-26
Prevents the slideshow interval to be set to 0 in the UI

### LINKED ISSUE #27: Slideshow delay allows zero delay value
lximageallows a slidshow delay of zero to be set in the prefrences which doesn't really make a sensible slideshow that changes pictures and instead ends up using 100 percent cpu for a single core switching the images as fast as the cpu goes which could lead to someone thinking they are clever than quickly running out of battery on a laptop or overheating there system maybe. I think a fix might be to take the maximum of 1 and  settings.slideShowInterval() in lximage-qt/src/prefrences-dialog on line 40. Or someone presses a key and hits 0 on accident and starts the slideshow would result in a system using 100 percent cpu.

npmiller:
Fixes #27

### DIFF
--- a/src/preferencesdialog.ui
+++ b/src/preferencesdialog.ui
@@ -73,7 +73,11 @@
         </widget>
        </item>
        <item row="3" column="1">
-        <widget class="QSpinBox" name="slideShowInterval"/>
+        <widget class="QSpinBox" name="slideShowInterval">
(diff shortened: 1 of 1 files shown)


luis-pereira:
@npmiller Thanks.

PR #30, opened 2015-07-23
Replace glib with Qt for command-line parsing

mharrys:
The comment in the code claim that there was no good way to use Qt for command-line parsing, but I found the solution provided by Qt to be much more clear and more appropriate. Using FM is also no longer needed in this function.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -58,10 +58,10 @@ bool Application::init(int argc, char** argv) {
     // we successfully registered the service
     isPrimaryInstance = true;
     setQuitOnLastWindowClosed(false); // do not quit even when there're no windows
-    
+
     new ApplicationAdaptor(this);
     dbus.registerObject("/Application", this);
-    
+
     connect(this, SIGNAL(aboutToQuit()), SLOT(onAboutToQuit()));
 
     if(settings_.useFallbackIconTheme())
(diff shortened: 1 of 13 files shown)


jleclanche:
lgtm, can you squash the two commits please? (`git rebase -i HEAD~2`, fixup the 2nd commit into the first one, `git push -f`)

luis-pereira:
Commit https://github.com/mharrys/lximage-qt/commit/e4be0772907cc31a003a8f379f579a0d9f4a6d06:
* Remove src/application.cpp:83 trailing space
* src/application.cpp:90 make the `QCommandLineParser::addPositionalArgument() syntax` argument also translatable.

mharrys:
Thanks for fast reply. I instead did a reset in my repo and also updated with the changes asked by @luis-pereira and then did a `git push -f`.

I had to turn off my script for removing trailing white space since the file had at least 8 cases of trailing white spaces before I started. If you want, I could create another PR for clearing out white spaces in the project.

luis-pereira:
@mharrys You can push the remove trailing spaces to this branch or place another PR. Whatever you prefer.

paulolieuthier:
GTM.

mharrys:
The trailing spaces for files in src directory has been removed. I also found a case where tabs were used instead of spaces.

luis-pereira:
Thanks @mharrys

PR #32, opened 2015-08-23
Add Greek (el) translation

dglent:
I added the Greek (el) translation for lximage
Files:
src/translations/lximage-qt-screenshot_el.desktop
src/translations/lximage-qt_el.desktop
src/translations/lximage-qt_el.ts

### DIFF
--- a/src/translations/lximage-qt-screenshot_el.desktop
+++ b/src/translations/lximage-qt-screenshot_el.desktop
@@ -1,0 +1,4 @@
+#Translations
+Name[el]=Στιγμιότυπο οθόνης
+GenericName[el]=Στιγμιότυπο οθόνης
+Comment[el]=Λήψη ενός στιγμιότυπου της οθόνης
--- a/src/translations/lximage-qt_el.desktop
+++ b/src/translations/lximage-qt_el.desktop
@@ -1,0 +1,4 @@
+#Translations
+Name[el]=LXImage
+GenericName[el]=Προβολέας εικόνων
+Comment[el]=Ο προβολέας εικόνων του LXQt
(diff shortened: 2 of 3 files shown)


pmattern:
Both filenames and content are stating language-specific locales ("el") only while line 3 in lximage-qt_el.ts is stating a country-specific one ("el_*GR*"). Shouldn't the latter be adjusted, too?

Assuming that country-specific translation, which is a recurring topic in general, isn't needed in Greek at all. But this applies, doesn't it?

dglent:
Yes it is truth that i didn't see a variant where it need different translations.
I have seen the Chyprus variant el_CY but i've never seen a discussion that there are different terms to describe the same thing as for exemple in pt and pt_BR
So el or el_GR it is the same but is better to be consistent.

What i have to do? commit again the same with --force ? is something else that i should make attention ?

pmattern:
Should be alright to limit the Greek translations to the language-specific "el" and omit any country-specific "el_FOO" then.

> What i have to do?

Correct file lximage-qt_el.ts, include that change in the recent commit by running `git commit --amend` and push basically the same way as with dglent/lximage-qt@22f3f6ab4b but add switch `--force`.
To my knowledge this PR will automatically get updated that way.

dglent:
The first commit i send it as pull request by the github site.
Now i couldn't to the same and i send it directly from my system.
Is it ok ?

[dglent@localhost lximage-qt (greek_translation)]$ git push --force
Username for 'https://github.com': dglent
Password for 'https://dglent@github.com': 
Counting objects: 7, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (7/7), done.
Writing objects: 100% (7/7), 3.12 KiB | 0 bytes/s, done.
Total 7 (delta 3), reused 0 (delta 0)
To https://github.com/dglent/lximage-qt.git
 + 22f3f6a...d7f0e13 greek_translation -> greek_translation (forced update)

dglent:
It is ok now with the second commit

palinek:
landed in @c4731e3, thanks

PR #41, opened 2015-12-21
Fix memory leak if taking screenshot with cursor

philippwiesemann:
There were two local variables with the same name in different scopes. The second variable stored the pointer but only the first was deleted.

Thanks!

### DIFF
--- a/src/screenshotdialog.cpp
+++ b/src/screenshotdialog.cpp
@@ -150,7 +150,6 @@ void ScreenshotDialog::doScreenshot() {
     if(cursor) {
       if(cursor->pixels) { // pixles should be an ARGB array
         QImage cursorImage;
-        quint32* buf = NULL;
         if(sizeof(long) == 4) {
           // FIXME: will we encounter byte-order problems here?
           cursorImage = QImage((uchar*)cursor->pixels, cursor->width, cursor->height, QImage::Format_ARGB32);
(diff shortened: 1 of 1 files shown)


src/screenshotdialog.cpp
```diff
@@ -157,7 +157,7 @@ void ScreenshotDialog::doScreenshot() {
         }
         else { // XFixes returns long integers which is not 32 bit on 64 bit systems.
           long len = cursor->width * cursor->height;
-          quint32* buf = new quint32[len];
+          buf = new quint32[len];
           for(long i = 0; i < len; ++i)
             buf[i] = (quint32)cursor->pixels[i];
           cursorImage = QImage((uchar*)buf, cursor->width, cursor->height, QImage::Format_ARGB32);
```
f2404: I think it's better to delete [] buf within this block, and remove the 'buf' declaration from line 153.

paulolieuthier: I agree.

philippwiesemann: `buf` can not be deleted there because `QImage()` does not copy it.

f2404: You mean, buf should be kept while cursorImage is in use?

palinek:
IMO all such code (new/delete) should be modified to use smart pointers (std's or Qt's).

In general I think, whereever we see a plain *delete*, we should consider using smart pointers...

philippwiesemann:
The commit was updated to use a different solution which also should be more correct.

paulolieuthier:
GTM.

f2404:
Looks good.

PR #42, opened 2015-12-25
Fix missing Russian translation in desktop file

philippwiesemann:
The Russian translation for the desktop entry was not included into the final desktop file because its keys did not end with the locale postfix.

Thanks!


f2404:
Yep, missed that. Thanks!

paulolieuthier:
Thanks!

Issue #43, opened 2015-12-27
Feature request: Option for the application window to resize automatically based on image dimensions


ms6676749:
Archlinux
lxqt 0.10.0
Xfwm4 4.12.3
lximage-qt 0.4.0

For example, open an image and lximage-qt will open it centerscreen in its native dimension or, if the image is too large, fit the image to the width or height of the screen accordingly.  Currently, lximage opens every image in a 553x432 window, which is pretty small on a 1920x1080 screen.

tsujan:
closed by #45

PR #45, opened 2016-01-03
Implement an EOG-like behavior on starting

tsujan:
By clicking on an image for the first time, the user wants to see it clearly with lximage-qt. So, the following behavior is implemented here:

(1) A minimum size of 400x400 is assumed;
(2) The window is scaled to fit the image;
(3) For too big images, the window is scaled down;
(4) The window is centered on the screen.
(5) The loaded image is selected and centered in the thumbnail view.

This also addresses #43.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -154,7 +154,9 @@ void Application::newWindow(QStringList files) {
       if(settings_.windowMaximized())
         window->setWindowState(window->windowState() | Qt::WindowMaximized);
 
-      window->show();
+      /* when there's an image, we show the window AFTER resizing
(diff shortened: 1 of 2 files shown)


jleclanche:
gtm, sorry for the delay

paulolieuthier:
I'm not sure about starting the windows always centered, but GTM.

tsujan:
> I'm not sure about starting the windows always centered

The user has no intention other than seeing the image when he clicks on it; hence centering the window. It could be made optional later -- lximage-qt still needs polishing.

PCMan:
Works as advertised. Thank you.

PR #47, opened 2016-01-14
Copy File Name

Pr0Wolf29:
Hello, I would like to contribute a feature I made for my own personal usage. I added a button to the toolbar and a right-click menu item that copies the file path of the currently opened image to the clipboard. I know it seems simple, but this is a huge help for me, especially when going through thousands of images for sorting purposes. I don't know much about libfm or Qt, but I made something that works, and followed the contribution guide as best I could within my memory's limits. I use this myself, so I suppose others might find it useful too.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -104,6 +104,8 @@ MainWindow::MainWindow():
   contextMenu_->addAction(ui.actionFlipHorizontal);
   contextMenu_->addAction(ui.actionFlipVertical);
   contextMenu_->addAction(ui.actionFlipVertical);
+  contextMenu_->addSeparator();
+  contextMenu_->addAction(ui.actionCopyFileName);
 
   // create shortcuts
   QShortcut* shortcut = new QShortcut(Qt::Key_Left, this);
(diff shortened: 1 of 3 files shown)


tsujan:
Thank you for taking time and trying to make LXImage-qt better!

LXImage-qt has already a "File Properties" menu-item, that can give you more info (to copy). IMHO, the usage you implemented seems too specific to your needs.

Pr0Wolf29:
Well if you say so. Otherwise I'll just have my own fork I'll maintain.

tsujan:
> Well if you say so...

I didn't close this PR but told my opinion. Other devs might have a different opinion.

pmattern:
The new feature proposed by this PR is to copy an image's *path* as a whole at once which isn't possible right now. While I didn't miss this feature personally so far I can for sure figure it's something that makes sense. E. g. [nomacs](http://nomacs.org) is providing this feature as well ("Copy" / `Ctrl+C` from images' context menus).
So I for one am +1 on this.

Not sure whether the current implementation makes really sense, though.
Naming "Copy *File Name*" needs to get changed for sure. It exactly is *not* the file name that gets copied, this is indeed working already as @tsujan rightly stated, but the path.
But that aside I wonder whether some other UI component may eventually make sense. E. g. maybe something like an address bar similar to the one of PCManFM-Qt or web browsers where the path could eventually be selected and copied as usual?

Pr0Wolf29:
Yeah, changing "file name" to "file path" would be a good idea.

PCMan:
The PR is out of date due to the recent libfm-qt changes. I'll take some time to fix it.

tsujan:
Superseded by https://github.com/lxqt/lximage-qt/pull/316

PR #48, opened 2016-01-23
Hide cursor smartly in fullscreen mode

tsujan:
This commit hides the cursor in the fullscreen mode after 3 seconds of inactivity and shows it again when needed.

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -29,6 +29,8 @@
 #include <QLabel>
 #include <QGraphicsProxyWidget>
 
+#define CURSOR_HIDE_DELY 3000
+
 namespace LxImage {
 
 ImageView::ImageView(QWidget* parent):
@@ -38,6 +40,7 @@ ImageView::ImageView(QWidget* parent):
   gifMovie_(NULL),
   autoZoomFit_(false),
   cacheTimer_(NULL),
+  cursorTimer_(NULL),
   scaleFactor_(1.0) {
 
   setViewportMargins(0, 0, 0, 0);
@@ -57,6 +60,10 @@ ImageView::~ImageView() {
   if(cacheTimer_) {
     cacheTimer_->stop();
     delete cacheTimer_;
+  }
+  if(cursorTimer_) {
+    cursorTimer_->stop();
+    delete cursorTimer_;
   }
 }
 
@@ -86,6 +93,34 @@ void ImageView::mouseDoubleClickEvent(QMouseEvent* event) {
   // filtered by event filter installed on the view.
   // QGraphicsView::mouseDoubleClickEvent(event);
   QAbstractScrollArea::mouseDoubleClickEvent(event);
+}
+
+void ImageView::mousePressEvent(QMouseEvent * event) {
+  QGraphicsView::mousePressEvent(event);
+  if(cursorTimer_) cursorTimer_->stop();
+}
+
+void ImageView::mouseReleaseEvent(QMouseEvent* event) {
+  QGraphicsView::mouseReleaseEvent(event);
+  if(cursorTimer_) cursorTimer_->start(CURSOR_HIDE_DELY);
+}
+
+void ImageView::mouseMoveEvent(QMouseEvent* event) {
+  QGraphicsView::mouseMoveEvent(event);
+   if(cursorTimer_ && (viewport()->cursor().shape() == Qt::BlankCursor
+                       || viewport()->cursor().shape() == Qt::OpenHandCursor)) {
+    cursorTimer_->start(CURSOR_HIDE_DELY); // restart timer
+    viewport()->setCursor(Qt::OpenHandCursor);
+  }
+}
+
+void ImageView::focusInEvent(QFocusEvent* event) {
+  QGraphicsView::focusInEvent(event);
+  if(cursorTimer_ && (viewport()->cursor().shape() == Qt::BlankCursor
+                      || viewport()->cursor().shape() == Qt::OpenHandCursor)) {
+    cursorTimer_->start(CURSOR_HIDE_DELY); // restart timer
+    viewport()->setCursor(Qt::OpenHandCursor);
+  }
 }
 
 void ImageView::resizeEvent(QResizeEvent* event) {
@@ -317,5 +352,26 @@ QRect ImageView::sceneToViewport(const QRectF& rect) {
   return QRect(topLeft, bottomRight);
 }
 
+void ImageView::blankCursor() {
+    viewport()->setCursor(Qt::BlankCursor);
+}
+
+void ImageView::hideCursor(bool enable) {
+  if(enable) {
+    if(cursorTimer_) delete cursorTimer_;
+    cursorTimer_ = new QTimer(this);
+    cursorTimer_->setSingleShot(true);
+    connect(cursorTimer_, SIGNAL(timeout()), this, SLOT(blankCursor()));
+    if(viewport()->cursor().shape() == Qt::OpenHandCursor)
+        cursorTimer_->start(CURSOR_HIDE_DELY);
+  }
+  else if (cursorTimer_) {
+    cursorTimer_->stop();
+    delete cursorTimer_;
+    cursorTimer_ = NULL;
+    if(viewport()->cursor().shape() == Qt::BlankCursor)
+        viewport()->setCursor(Qt::OpenHandCursor);
+  }
+}
 
 } // namespace LxImage
--- a/src/imageview.h
+++ b/src/imageview.h
@@ -67,9 +67,16 @@ public:
     autoZoomFit_ = value;
   }
 
+  // if set to true, hides the cursor after 3s of inactivity
+  void hideCursor(bool enable);
+
 protected:
   virtual void wheelEvent(QWheelEvent* event);
   virtual void mouseDoubleClickEvent(QMouseEvent* event);
+  virtual void mousePressEvent(QMouseEvent* event);
+  virtual void mouseReleaseEvent(QMouseEvent* event);
+  virtual void mouseMoveEvent(QMouseEvent* event);
+  virtual void focusInEvent(QFocusEvent* event);
   virtual void resizeEvent(QResizeEvent* event);
   virtual void paintEvent(QPaintEvent* event);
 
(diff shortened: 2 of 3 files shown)


src/imageview.cpp
```diff
@@ -317,5 +346,26 @@ QRect ImageView::sceneToViewport(const QRectF& rect) {
...
 }
 
+void ImageView::blankCursor() {
+    viewport()->setCursor(Qt::BlankCursor);
+}
+
+void ImageView::hideCursor(bool enable) {
+  if(enable) {
+    if(cursorTimer_) delete cursorTimer_;
+    cursorTimer_ = new QTimer(this);
```
palinek: Wouldn't it be better to just create/initialize the `cursorTimer_` object only in case not existing, t.m.:
```c++
if (!cursorTimer_)
{
  cursorTimer_ = new QTimer(this);
...
```

...and second question: Does a `QTimer` object use much resources -> why not to create/initialize it just once in the `ImageView` constructor and use it later only when needed?

tsujan:
By deleting the timer, its signal will be disconnected if not triggered yet. Yes, I could disconnect it explicitly but it's just a single-shot timer. No practical difference -- just a matter of taste or style.

> why not to create/initialize it just once

Simply because this way seems more logical to me; it isn't always needed -- just when the window is in the full-screen mode. PCMan has done the same thing with `cacheTimer_`. However, no tangible difference and only a matter of style, again ;)

This was a very simple commit. Its only "tricky" part was the need to focusInEvent(), which didn't seem needed at first (full-screen -> change desktop -> return to the same desktop).

palinek:
> By deleting the timer, its signal will be disconnected if not triggered yet. Yes, I could disconnect it explicitly but it's just a single-shot timer.

SingleShot doesn't mean it can emit the signal only once in the object lifetime. IMO all the recreation/reinitialization here (in case the timer object exists) is unnecessary.

tsujan:
@palinek  I really don't get your point. If my method seems weird to you, feel free to change it into your way. As far as it works too, it's completely fine to me.

On the *practical* side, do you see any problem? Doesn't it work correctly for you in some situation? Which one and how?

palinek:
> On the practical side, do you see any problem? Doesn't it work correctly for you in some situation? Which one and how?

Nope. Just the small optimization for not deleting & reinitialization of `QTimer` object. Something like this:
```diff
diff --git a/src/imageview.cpp b/src/imageview.cpp
index e791126..65d6b3c 100644
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -38,7 +38,6 @@ ImageView::ImageView(QWidget* parent):
   gifMovie_(NULL),
   autoZoomFit_(false),
   cacheTimer_(NULL),
-  cursorTimer_(NULL),
   scaleFactor_(1.0) {
 
   setViewportMargins(0, 0, 0, 0);
@@ -352,17 +351,17 @@ void ImageView::blankCursor() {
 
 void ImageView::hideCursor(bool enable) {
   if(enable) {
-    if(cursorTimer_) delete cursorTimer_;
-    cursorTimer_ = new QTimer(this);
-    cursorTimer_->setSingleShot(true);
-    connect(cursorTimer_, SIGNAL(timeout()), this, SLOT(blankCursor()));
+    if (!cursorTimer_)
+    {
+      cursorTimer_.reset(new QTimer);
+      cursorTimer_->setSingleShot(true);
+      connect(cursorTimer_.data(), SIGNAL(timeout()), this, SLOT(blankCursor()));
+    }
     if(viewport()->cursor().shape() == Qt::OpenHandCursor)
         cursorTimer_->start(3000);
   }
   else if (cursorTimer_) {
-    cursorTimer_->stop();
-    delete cursorTimer_;
-    cursorTimer_ = NULL;
+    cursorTimer_.reset(nullptr);
     if(viewport()->cursor().shape() == Qt::BlankCursor)
         viewport()->setCursor(Qt::OpenHandCursor);
   }
diff --git a/src/imageview.h b/src/imageview.h
index ca4a7bd..d334aa0 100644
--- a/src/imageview.h
+++ b/src/imageview.h
@@ -98,7 +98,7 @@ private:
   QRect cachedRect_; // rectangle containing the cached region (in viewport coordinate)
   QRect cachedSceneRect_; // rectangle containing the cached region (in scene/original image coordinate)
   QTimer* cacheTimer_;
-  QTimer *cursorTimer_; // for hiding cursor in fullscreen mode
+  QScopedPointer<QTimer> cursorTimer_; // for hiding cursor in fullscreen mode
   double scaleFactor_;
   bool autoZoomFit_;
 };
```

tsujan:
I see no (need for an) optimization here. QScopedPointer isn't really needed for such a trivial job.

First, the line `if(cursorTimer_) delete cursorTimer_;` is only there to respect the logic; it has no effect because the `if` part is never true.

Second, there isn't ANY problem with deleting a QTimer when it isn't needed. As I mentioned before, PCMan has done exactly the same thing with `cacheTimer_` in `ImageView::queueGenerateCache()`. I don't mean it's correct because *he* has done so but, because it has *no* problem, both of us have naturally done so.

You've mistaken a difference in style for a real problem in this special case. However, thank you for taking time.

BTW, I added another deleting in the destructor too, as PCMan had done (not needed really) ;)

palinek:
> I see no (need for an) optimization here. QScopedPointer isn't really needed for such a trivial job.

Many wise guys (including Bjarne Stroustrup) strongly encourage to use *smart pointers* and to paraphrase them: "...whenever you see a naked *delete* in the code, be suspicious...". But the smart pointer wasn't the point...

The point was the superflous initialization. Compare this:
```c++
  if(cursorTimer_) delete cursorTimer_;
  cursorTimer_ = new QTimer(this);
  cursorTimer_->setSingleShot(true);
  connect(cursorTimer_, SIGNAL(timeout()), this, SLOT(blankCursor()));
```
with something like this:
```c++
  if(!cursorTimer_)
  {
    cursorTimer_ = new QTimer(this);
    cursorTimer_->setSingleShot(true);
    connect(cursorTimer_, SIGNAL(timeout()), this, SLOT(blankCursor()));
  }
```

> First, the line `if(cursorTimer_) delete cursorTimer_;` is only there to respect the logic; it has no effect because the `if` part is never true.

Then... 1. the code is a bit misleading.  2. it is a branch, so a potential performance hit.
=> IMO the latter (`if(!cursorTimer_)...`) is better readable.

If this doesn't make any sense... then forget anything I've said here before.

tsujan:
@palinek 
> Many wise guys (including Bjarne Stroustrup) strongly encourage...

Then, neither I nor PCMan follows what they say -- perhaps each of us for his own reason. No trace of QScopedPointer in pcmanfm, libfm-qt, lximage-qt,.... Even if I had the same opinion as you do, I wouldn't go with it because I wouldn't want to change the original author's coding style.

> The point was the superflous initialization

And I could say exactly the same thing about `if(!cursorTimer_)` in your suggestion because it's always `true`!

> the code is a bit misleading

To whom? Of course you meant yourself but that's a matter of perspective.

> it is a branch, so a potential performance hit

Which performance hit?!

Is it so hard to allow for a simple difference in style/taste while everything works OK? I can easily accept that you like to follow principles while I'm a pragmatist. Both are OK to me -- but let's not start a philosophical discussions *here*!

palinek:
> Both are OK to me -- but let's not start a philosophical discussions here!

I don't want any flame here too...

PCMan:
Well, please avoid flame war if possible. To me, it's more a coding style issue than a performance issue. To be honest, my C++ coding style is not good. I never used smart pointers simply because I used to handle the pointers myself and the code was mostly written prior to the widespread adoption of C++ 11.
(It's generally good practice to use it whenever possible as suggested by the father of C++ and I agree. But if you can handle raw pointers correctly, not using smart pointer is not a sin.)
As a gtk+ user for 10 years, I learned how to use Qt when I started to work on LXQt.
So it's my first time using Qt. So my coding style might not be a good reference. :-)
Allocating an object and deleting it may take hundreds of CPU cycles, but there's no visible performance difference. A memory cache miss takes much more than that. We don't need to do micro-optimizations.
Please use whatever you think is more readable unless it's proven to be a performance bottle neck.
I'm OK with any coding style or good practice suggested by c++ experts.

tsujan:
> Well, please avoid flame war if possible.

@palinek  and I implicitly agreed on that :)

> To be honest, my C++ coding style is not good.

LOL! Your coding style is one of the cleanest I've ever seen.

> I never used smart pointers simply because I used to handle the pointers myself...

Almost similarly, I like to have full control over allocating and deleting.

paulolieuthier:
My two cents: I don't like `QScopedPointer`, but I like to avoid using `delete` as well. I'd simply not use a pointer at all, and call `start()` and `stop()`.

src/imageview.cpp
```diff
@@ -88,6 +93,34 @@ void ImageView::mouseDoubleClickEvent(QMouseEvent* event) {
...
 }
 
+void ImageView::mousePressEvent(QMouseEvent * event) {
+  QGraphicsView::mousePressEvent(event);
+  if(cursorTimer_) cursorTimer_->stop();
+}
+
+void ImageView::mouseReleaseEvent(QMouseEvent* event) {
+  QGraphicsView::mouseReleaseEvent(event);
+  if(cursorTimer_) cursorTimer_->start(3000);
```
philippwiesemann: If there would be a constant for the milliseconds then only one place would need an update if the value should be changed later. :)

tsujan: You're right.

tsujan:
@paulolieuthier 
As I repeatedly said here and there, IMO, tastes are different and all of them are acceptable more or less. I just can't understand why such a hot and long debate about such a simple change. Yes, `delete` and `NULL` should be handled with care but that doesn't mean they're taboos.

I wanted to work on https://github.com/lxde/lximage-qt/issues/49 *after* this PR because of common files. I'm getting the impression that lximage-qt may not be liked/used by anyone. Please tell me if that's true because I don't want to work on a "deprecated" app!

agaida:
@tsujan - it is used and i think it will become more important in the near futute

tsujan:
> it is used and i think it will become more important in the near futute

Then let its users enjoy this simple feature. Anyone could make a PR later. And I want to work on SVG resizing and other features.

PR #60, opened 2016-08-03
Quieten compiler warning

tsujan:
Thanks to, https://github.com/lxde/lximage-qt/pull/58/commits/8c52c8ddf0a14369d6b0f53e7672992294d2d287, a problem is also found and fixed.

A warning related to `g_io_scheduler_job_send_to_mainloop()` remains. I think @PCMan might fix it better than I could.

### DIFF
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -27,11 +27,11 @@ using namespace LxImage;
 Settings::Settings():
   useFallbackIconTheme_(QIcon::themeName().isEmpty() || QIcon::themeName() == "hicolor"),
   bgColor_(255, 255, 255),
+  fullScreenBgColor_(0, 0, 0),
   showThumbnails_(false),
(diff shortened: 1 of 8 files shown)


src/settings.cpp
```diff
@@ -49,7 +49,7 @@ bool Settings::load() {
   fullScreenBgColor_ = settings.value("fullScreenBgColor", fullScreenBgColor_).value<QColor>();
   // showThumbnails_;
   // showSidePane_;
-  int slideShowInterval_ = settings.value("slideShowInterval", slideShowInterval_).toInt();
+  slideShowInterval_ = settings.value("slideShowInterval", slideShowInterval_).toInt();
```
tsujan: This was a hidden bug.

luis-pereira:
GTM

PR #64, opened 2016-08-24
Remove Core and Qt from Categories in desktop file

jubalh:
I don't see what they are needed for. Am I missing anything here?


pmattern:
Agree. One of the ["main categories"](https://specifications.freedesktop.org/menu-spec/latest/apa.html#main-category-registry) persists so this won't make `desktop-file-validate` complain either.

GTM.

agaida:
GTM

paulolieuthier:
GTM.

PR #65, opened 2016-09-04
changed timeout for image cache rebuilding timer from 200ms to immediate

aldn:
Now the downsampled image pops immediately, without displaying the coarse version for the first 200ms. Not sure why delay was put there in the first place.

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -322,7 +322,7 @@ void ImageView::queueGenerateCache() {
     connect(cacheTimer_, &QTimer::timeout, this, &ImageView::generateCache);
   }
   if(cacheTimer_)
-    cacheTimer_->start(200); // restart the timer
+    cacheTimer_->start(); // restart the timer
 }
 
 // really generate the cache


aldn:
Hello?

tsujan:
Sorry for the late response! I was very busy.

Qt doc says, "A QTimer with a timeout interval of 0 will time out as soon as all the events in the window system's event queue have been processed." So, I think @PCMan set the timeout to 200ms for certainty.

What do you exactly mean by "Now the downsampled image pops immediately"? I can't see any delay here.

aldn:
>  I can't see any delay here.
Try opening an image larger than current window size. The first thing I see when I select the image is the nearest-neighbor-like scaled image. After 200 ms the "cache" image is built (with flag Qt::SmoothTransformation) and is used since then.  You can change 200 to other value to exagerrate the effect described.

>set the timeout to 200ms for certainty.
Certainty for what ? Why we'd want to delay the processing of cached image by 200 ms for any reason ?

PCMan:
@aldn The reason for the delay is quite simple. To avoid unnecessary computation.
The smooth transformation can be very slow on large images, especially if you have an older machine.
If you try to zoom in or zoom out the image very quickly, without the delay, the smooth transformation might be done for many times. If you queue all of them and delay the handling, you only need to do it once.
The other use case is, when you zoom in the image to a very large scale, and try to do panning, the smooth version of the visible region needs to be generated on the fly, and we don't want to do it too often.
Another simple case is when you try to switch to the next image repeatedly, without the delayed processing, the smoothing calculation will be done many times, which is obviously unnecessary.
So, please don't apply the patch.
The timer is there by design, for faster UI response.

aldn:
With panning - I agree with you, the delay would help to avoid situation like rebuilding the cache  only to invalidate it shortly after.Maybe the delay should be added only in certain cases (with panning as you mentioned).  But for normal browsing the effect is highly deterring. 
I mean, when I "scrollwheel" my way through a large collection of images I usually stay for a while on each one. From the user experience standpoint the need to wait for smooth version to be loaded is quite detrimental, especially since CPU is doing nothing for good 200 ms.

tsujan:
@PCMan, thanks for the explanation! Would you please add a comment to the code?

tsujan:
> From the user experience standpoint the need to wait for smooth version to be loaded is quite
detrimental, especially since CPU is doing nothing for good 200 ms.

I agree with you on this.

tsujan:
Another method should be devised to remove the delay only with browsing (thanks to @aldn for mentioning that). The delay is definitely needed with rescaling. So, I close the PR.

PR #71, opened 2016-10-01
Fixed A Cast

tsujan:
Corrected a wrong cast that deformed (scaled) images on rotation. (A bad typo plus a wrong assumption? I don't remember.)


tsujan:
I don't know about releases but if the new LXQt isn't released yet, please test this PR and merge it if OK!

PR #72, opened 2016-10-06
[MRG] Display first image when loading a directory

asnt:
This PR addresses https://github.com/lxde/lxqt/issues/1117.

Goal: When loading a directory from CLI or from a dialog, display the first image of the directory.
- [x] Load directory from CLI.
- [x] Load directory from Dialog.

The current implementation uses QDir for listing the files in a directory. Is it OK with the project policy? Or should one use libfm-qt functions where possible?

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -345,7 +345,7 @@ void MainWindow::on_actionOpenFile_triggered() {
 void MainWindow::on_actionOpenDirectory_triggered() {
   QString directory = openDirectory();
   if(!directory.isEmpty()) {
-    openImageFile(directory);
+    openImageDirectory(directory);
   }
 }
 
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -19,6 +19,8 @@
 
 
 #include "mainwindow.h"
+#include <algorithm>
+#include <QCollator>
 #include <QDir>
 #include <QFileInfo>
 #include <QMessageBox>
@@ -200,7 +202,7 @@ void MainWindow::openImageFile(QString fileName) {
                                                       qPrintable(fileName),
                                                       nullptr);
   if (info.isDir()) {
-      QString imageString = firstImageOfDir(fileName);
+      QString imageString = findFirstImageOfDir(fileName);
       FmPath* imagePath = fm_path_new_for_str(qPrintable(imageString));
       loadImage(imagePath);
       fm_path_unref(imagePath);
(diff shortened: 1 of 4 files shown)


src/mainwindow.cpp
```diff
@@ -193,9 +196,20 @@ void MainWindow::openImageFile(QString fileName) {
     fm_path_unref(path);
     return;
   }
-  // load the image file asynchronously
-  loadImage(path);
-  loadFolder(fm_path_get_parent(path));
+  Fm::FileInfo info = Fm::FileInfo::newFromNativeFile(nullptr,
+                                                      qPrintable(fileName),
+                                                      nullptr);
+  if (info.isDir()) {
```
tsujan: Isn't this simpler?
````
if (QFileInfo (fileName).isDir()) {
````

File: src/mainwindow.cpp [hunk 2]
```diff
@@ -268,13 +288,38 @@ QString MainWindow::saveFileName(QString defaultName) {
...
+    dir.setNameFilters(formats);
+    dir.setFilter(QDir::Files | QDir::NoDotAndDotDot);
+    dir.setSorting(QDir::Name);
+    QList<QFileInfo> imageFiles = dir.entryInfoList();
+    if (!imageFiles.empty()) {
+        return imageFiles.first().absoluteFilePath();
+    }
+    return QString();
+}
+
```
tsujan: Unfortunately, `MainWindow::firstImageOfDir()` doesn't necessarily give the first image that can be seen in the thumbnail view. For example, if a folder contains "3.jpg", "7.jpg", "101.jpg" and "411.jpg", this function gives  "101.jpg" while "3.jpg" is the first image in the thumbnail view.

src/mainwindow.cpp
asnt: Thank you for your review @tsujan .
Yes, it is indeed simpler. Hence my questions in the first message. What is the policy regarding the use of Qt functions versus libfm-qt functions where both provide similar functionalities?

tsujan:
> What is the policy regarding the use of Qt functions versus libfm-qt functions where both provide similar functionalities?

Personally, I prefer to use Qt methods in Qt apps. `libfm-qt` is for file management. For example, the thumbnail view in lximage-qt is `Fm::FolderView*`. It could be said, "why not using Qt-only methods for file management?" The answer is that we already have `libfm`, so we have no reason to try to do the complex job again in Qt. Moreover, a Qt-only library may not be possible with the same functionalities.

asnt:
@tsujan That makes sense. Thank you for your clarification.

tsujan:
@asnt 
Thank you for your contribution. It seems your last commits work.

However, we can use the current functions to do the job more easily. For example, please see this patch: https://dl.dropboxusercontent.com/u/7793460/lximage-qt_open-folder.diff.gz

PR #75, opened 2016-10-26
[MRG] Display first image when loading a directory (v2)

asnt:
Display the first available image when opening a folder. Open a folder from
CLI and from a dialog, via the menu or Ctrl+D.

Thanks @tsujan for the reviews and suggestions.

Address lxde/lxqt#1117
Improve upon #72

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -187,7 +187,7 @@ void MainWindow::onFolderLoaded(FmFolder* folder) {
     }
   }
   // this is used to open the first image of a folder
-  if (!currentFile_)
+  else if (currentFile_ == nullptr)
     on_actionFirst_triggered();
 }
 
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -199,7 +199,18 @@ void MainWindow::openImageFile(QString fileName) {
     return;
   }
   if (QFileInfo(fileName).isDir()) {
-    if(fm_path_equal(path, folderPath_) || !dirHasImage(fileName)) {
+    if(fm_path_equal(path, folderPath_)) {
+      fm_path_unref(path);
+      return;
+    }
+    QList<QByteArray> formats = QImageReader::supportedImageFormats();
+    QStringList formatsFilters;
+    for (const QByteArray& format: formats)
+      formatsFilters << QString("*.") + format;
+    QDir dir(fileName);
+    dir.setNameFilters(formatsFilters);
+    dir.setFilter(QDir::Files | QDir::NoDotAndDotDot);
+    if(dir.entryList().isEmpty()) {
       fm_path_unref(path);
       return;
     }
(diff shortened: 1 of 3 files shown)


src/mainwindow.cpp
```diff
@@ -184,6 +186,9 @@ void MainWindow::onFolderLoaded(FmFolder* folder) {
       thumbnailsView_->childView()->scrollTo(currentIndex_, QAbstractItemView::EnsureVisible);
     }
   }
+  // this is used to open the first image of a folder
+  if (!currentFile_)
```
tsujan: Please add an `else` here:
````
else if (!currentFile_)
````

File: src/mainwindow.cpp [hunk 2]
```diff
@@ -193,12 +198,35 @@ void MainWindow::openImageFile(QString fileName) {
...
+  QList<QByteArray> formats = QImageReader::supportedImageFormats();
+  QStringList formatsFilters;
+  for (const QByteArray& format: formats)
+    formatsFilters << QString("*.") + format;
+  QDir dir(dirname);
+  dir.setNameFilters(formatsFilters);
+  dir.setFilter(QDir::Files | QDir::NoDotAndDotDot);
+  return !dir.entryList().isEmpty();
+}
+
```
tsujan: MainWindow::dirHasImage() seems redundant to me because it doesn't use any class member. We can remove it (from the header too) and put its contents into openImageFile() as:
````
void MainWindow::openImageFile(QString fileName) {
  FmPath* path = fm_path_new_for_str(qPrintable(fileName));
  if(currentFile_ && fm_path_equal(currentFile_, path)) {
    // the same file! do not load it again
    fm_path_unref(path);
    return;
  }
  if (QFileInfo(fileName).isDir()) {
    if(fm_path_equal(path, folderPath_)) {
      fm_path_unref(path);
      return;
    }
    QList<QByteArray> formatsBytes = QImageReader::supportedImageFormats();
    QStringList formats;
    for (const QByteArray& formatBytes: formatsBytes)
      formats << QString("*.") + formatBytes;
    QDir dir(fileName);
    dir.setNameFilters(formats);
    dir.setFilter(QDir::Files | QDir::NoDotAndDotDot);
    if(dir.entryInfoList().isEmpty()) {
      fm_path_unref(path);
      return;
    }
    if(currentFile_)
      fm_path_unref(currentFile_);
    currentFile_ = nullptr;
    loadFolder(path);
  } else {
    // load the image file asynchronously
    loadImage(path);
    loadFolder(fm_path_get_parent(path));
  }
  fm_path_unref(path);
}
````
Please also correct the indentation (double space) as above and this PR will be ready for a merge.

tsujan:
With the above changes, this is better and more frugal than https://github.com/lxde/lximage-qt/pull/72, IMO.

src/mainwindow.cpp
asnt: I personnaly find it less readable to have all these operations packed into the same method.

tsujan:
GTM.

agaida:
so #72 is obsoleted?

asnt:
@agaida Indeed. Now closed.

tsujan:
@agaida, I use this PR since it's made and have encountered no problem. On the coding level, I already gave my "GTM".

PR #77, opened 2016-11-25
NotReady: add custom image rotation

rndevfx:
I've created a custom rotation possibility and within the menu you can find a new entry now: Edit -> Rotate Custom

Need to fix the problems:

- [ ] When you rotate the image +10 degrees and back -10 degrees for example the image is resized and you'll get an offset
- [ ] When you load a GIF (or SVG), 90deg rotation works fine and the image stays in centre. But if you custom-rotate it, there is an offset of the centre point. The current centre point calculation is too basic and a much better calculation must be used

Feedback or suggestions are very welcome 😄

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -26,6 +26,7 @@
 #include <QImageWriter>
 #include <QClipboard>
 #include <QPainter>
+#include <QInputDialog>
 #include <QPrintDialog>
 #include <QPrinter>
 #include <QWheelEvent>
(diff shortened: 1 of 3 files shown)


HybridDog:
As far as I known, image rotation by an angle which is not a multiple of 90° is non-trivial for raster graphics:
* The operation is lossy
* There are many options for how to calculate colours at integer positions from the rotated pixels which often have non-integer positions, e.g.:
  * The interpolation used, e.g. nearest neighbour, linear, cubic, lohalo, lanczos, etc. The most suitable one could depend on the image content and intended use of the image, and it may vary at different locations in the image.
  * Whether to rotate it gamma-correctly, which should work well for photos, perceptually, which could work well for screenshots of text, or in the colour space of the input image, which likely requires the lowest calculation time
  * Handling of fully-transparent and semi-transparent pixels
  * Calculation of output colours when the input image is not in sRGB 
* The image can be cropped or pixels which were out-of-bounds in the input image can be set to some value

Personally I prefer to use an image editor, such as the GIMP or vips, instead of an image viewer to edit images, so I'm not in favour of this PR.

PR #78, opened 2016-11-26
Fix linker error on FreeBSD >=10 (with Clang)

OlivierDuchateau:
I noticed, lximage-qt build fails with Clang (on FreeBSD > 10). Linker does not find -lexif even we explicite pass LDFLAGS variable.

Using EXIF_LDFLAGS (it adds all required linker flags) in target_link_libraries function fixes this issue.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -83,7 +83,7 @@ set(QT_LIBRARIES Qt5::Widgets Qt5::Core Qt5::DBus Qt5::PrintSupport Qt5::X11Extr
 target_link_libraries(lximage-qt
     fm-qt
     ${QT_LIBRARIES}
-    ${EXIF_LIBRARIES}
+    ${EXIF_LDFLAGS}
     ${X11_LIBRARIES}
     ${XFIXES_LIBRARIES}
 )


agaida:
> [100%] Linking CXX executable lximage-qt
> /usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/libfm.so: undefined reference to symbol 'exif_data_unref'
> //usr/lib/x86_64-linux-gnu/libexif.so.12: error adding symbols: DSO missing from command line
> collect2: error: ld returned 1 exit status
> src/CMakeFiles/lximage-qt.dir/build.make:643: die Regel für Ziel „src/lximage-qt“ scheiterte
> make[2]: *** [src/lximage-qt] Fehler 1
> CMakeFiles/Makefile2:86: die Regel für Ziel „src/CMakeFiles/lximage-qt.dir/all“ scheiterte
> make[1]: *** [src/CMakeFiles/lximage-qt.dir/all] Fehler 2
> Makefile:127: die Regel für Ziel „all“ scheiterte
Leads to a FTBFS in Linux - so i guess not the best of all ideas.

PCMan:
@OlivierDuchateau does this bug still exist? The EXIF lib is now handled by FindExif.cmake shipped with lxqt-build-tools, which I believe does not has the problem.

agaida:
Ping @OlivierDuchateau  - is this still valid and worked on? If not please close the PR.

BabylonAS:
I don’t recall any issue with EXIF when I was building LXQt on FreeBSD 11 myself, hence I didn’t have to change it to EXIF_LDFLAGS (it was XFIXES that required such a change, see https://github.com/lxqt/lximage-qt/pull/176). So It can be assumed this problem is out. But now I am worried if any change where LIBRARIES is replaced by LDFLAGS (I have submitted multiple such pull requests) adversely affects compilation on Linux in favor of FreeBSD...

PR #79, opened 2016-12-06
Fix typo in lximage-qt-screenshot_fr.desktop

paiiou:


pmattern:
Landed in 53635d6. Really wonder why I didn't stumble upon this before merging 1d14a66. Funny enough even `desktop-file-validate` whines about stuff totally irrelevant in this localization file but happily accepts the missing bracket. Anyway...

Merci.

PR #81, opened 2016-12-16
File and folder DND

tsujan:
Fixes https://github.com/lxde/lximage-qt/issues/69.

A file/folder can be dropped into lximage-qt's view and it will be opened if it is/contains a valid image.

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -35,7 +35,7 @@ namespace LxImage {
 
 ImageView::ImageView(QWidget* parent):
   QGraphicsView(parent),
-  scene_(new QGraphicsScene(this)),
+  scene_(new GraphicsScene(this)),
   imageItem_(new QGraphicsRectItem()),
   gifMovie_(nullptr),
   cacheTimer_(nullptr),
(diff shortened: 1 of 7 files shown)


src/imageview.cpp
```diff
@@ -49,6 +49,7 @@ ImageView::ImageView(QWidget* parent):
   setLineWidth(0);
 
   setScene(scene_);
+  connect(scene_, &GraphicsScene::fileDropped, this, &ImageView::onFileDropped);
```
palinek: Just curious, isn't there a way to connect SIGNAL->SIGNAL in the Qt5 notation?

tsujan: I don't know of any.

agaida:
I don't like it - i love it :D - GTM

tsujan:
@agaida 
LOL! Thanks!

I just don't know if it would be better to make `graphicsscene.h` a private header. @palinek, any idea?

tsujan:
Ok, standardized the code by adding "graphicsscene.cpp".

tsujan:
Anyone against merging this? Its logic is simple.

palinek:
Ship it.

PR #82, opened 2016-12-16
Bump patch version and

agaida:
build-tools version too - doesn't hurt, because pcmanfm-qt/libfm-qt need 0.3.0 too


tsujan:
GTM.

PR #97, opened 2017-06-27
set Qt::AA_UseHighDpiPixmaps to true

agaida:
refs lxde/lxqt/issues/916


agaida:
@tsujan - the rest of the hight-dpi patches are more or less the same, but here i'm not so sure, different notation. Could you have a look?

tsujan:
This is OK.

PR #100, opened 2017-07-05
Add Portuguese Brazil

brunobenet8:


tsimonq2:
:+1:

src/translations/lximage-qt-screenshot_pt_br.desktop
```diff
@@ -0,0 +1,4 @@
+#Translations
+Name[pt]=Captura de tela
```
palinek: All keys should be `...[pt_BR]=...`, not?

agaida:
@luis-pereira @palinek Only a hint - we can modify PRs if they are useful - just to speed up things a bit :sunglasses: I don't want to wait another year.

PR #103, opened 2017-07-16
Update loadimagejob.cpp

brunobenet8:
Lack of parentheses...

### DIFF
--- a/src/loadimagejob.cpp
+++ b/src/loadimagejob.cpp
@@ -58,14 +58,14 @@ void LoadImageJob::exec() {
       gssize readSize = g_input_stream_read(inputStream,
                                             buffer, 4096,
                                             cancellable().get(), &error);
-      if(readSize == -1 || readSize == 0) // error or EOF
+      if(readSize == -1 || readSize == 0){ // error or EOF
         break;
       // append the bytes read to the image buffer
         imageBuffer.append(buffer, readSize);
     }
     g_input_stream_close(inputStream, NULL, NULL);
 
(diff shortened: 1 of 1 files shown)


palinek:
1. with your patch the sources aren't compilable
2. why do you think, there is the brace missing?

brunobenet8:
Hi, well, has many parts of the code without brace, in the "if" never
starts, so i changed some parts, but still is incomplete.

Em 17/07/2017 3:07 AM, "palinek" <notifications@github.com> escreveu:


   1. with your patch the sources aren't compilable
   2. why do you think, there is the brace missing?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/lxde/lximage-qt/pull/103#issuecomment-315674302>, or mute
the thread

palinek:
This is simply wrong... you're trying to talk about the coding style, but with this PR you broke the app and it isn't even compilable.... so closing

tsujan:
@brunobenet8 We don't have an obsession with braces.

brunobenet8:
Hi, sorry that changes, was that i make a fork from project because i use
the lubuntu how my main system and i need that the image view send e-mail
from the file images, but in debugger never compile, so they show the
missing braces, so sorry that i changes before input the new code.

2017-07-19 6:54 GMT-03:00 tsujan <notifications@github.com>:

tsujan:
> but in debugger never compile,

Your system should have another problem. There's no missing braces in any LXQt source.

PR #104, opened 2017-07-16
Added Lithuanian .desktop file

welaq:


palinek:
landed in 914f09b, thanks

PR #106, opened 2017-07-19
Update loadimagejob.cpp

brunobenet8:
Some parts without brace...


palinek:
The very same ase #103

PR #108, opened 2017-07-20
Fix regression in thumbnail view

tsujan:
Fixes https://github.com/lxde/lximage-qt/issues/102.


agaida:
Works good

tsujan:
It's logic: after recent changes, the real size is given by `FolderItemDelegate`, not by the list view.

PR #111, opened 2017-09-02
Don't use hardcoded install dir

luis-pereira:
This way we don't make assumptions about install dirs. And we give the
packager the ability to override CMAKE_INSTALL_DATAROOTDIR.


agaida:
Full ack, GTM

PR #114, opened 2017-10-18
Bump versions

agaida:
Remove some QUIET


tsujan:
Yes, I also think `QUIET` should have been removed in those cases. GTM.

tsujan:
... Or the SVG dependency could be loosened but I see no practical reason for that.

agaida:
There is a bug for it - removing the QUIETs - if we really want some components optional there are better ways - like options.

palinek:
> if we really want some components optional there are better ways - like options.

QUIET isn't about making the dependency optional... but about not-writing information about found package during the build phase

palinek:
GTM

tsujan:
> QUIET isn't about making the dependency optional... but about not-writing information about found package during the build phase

... which can be used best with optional dependencies.

Anyhow, I don't think we need optional features here.

agaida:
@palinek  - right, i know - and i really hate this kind of CMakeFiles that work wit if found - and i really like the straight forward approach - find has to fail if a component that i want to have isn't there - if i use QUIET i can only hope that it will be checked/handled later.

Issue #98, opened 2017-06-30
Add some kind of upload functionality


agaida:
it would be nice if it would possible without hard-coding possible upload targets

tsujan:
Sorry if this is a stupid question -- I always upload things with browser

Could you please be more specific about how it should be? A general sketch?

agaida:
like in the former screengrab - a functionality to upload screenshots to imgur and other services

tsujan:
I need to look into the code of a Qt app that does the same thing or maybe someone who has done it before could add a PR.

agaida:
Maybe it can be done with some kind of bindings that don't require a recompile after adding an upload target - that would be really cool.

agaida:
erm - not that important, but very comfty :P

nathan-osman:
I submitted a pull request that achieves this and includes support for Imgur.

tsujan:
@nathan-osman Thanks for your contribution!

PR #115, opened 2017-10-20
Add support for uploading files

nathan-osman:
I decided to take a stab at bug #98.

The user interface I came up with is fairly simple.

1. The user clicks the "Edit->Upload" menu item and selects a provider:

    screenshot_20171020_000645

2. The image is uploaded asynchronously:

    screenshot_20171020_000701

3. Upon successful completion, the URL of the image is shown:

    screenshot_20171020_000720

Under the hood, I introduced a few new classes:

- UploadDialog takes care of presenting the UI to the user and displaying the status of the upload
- Provider is a base class for creating Upload objects, given a file opened for reading
- Upload is returned by Provider and emits signals to indicate upload status

This pull request includes support for Imgur uploads but the code is designed to easily allow more services to be supported.

### DIFF
--- a/src/upload/imgurprovider.h
+++ b/src/upload/imgurprovider.h
@@ -1,6 +1,6 @@
 /*
-    <one line to give the program's name and a brief idea of what it does.>
-    Copyright (C) 2013  PCMan <email>
+    LxImage - image viewer and screenshot tool for lxqt
+    Copyright (C) 2017  Nathan Osman <nathan@quickmediasolutions.com>
 
     This program is free software; you can redistribute it and/or modify
     it under the terms of the GNU General Public License as published by
--- a/src/upload/imgurprovider.h
+++ b/src/upload/imgurprovider.h
@@ -1,0 +1,47 @@
+/*
+    <one line to give the program's name and a brief idea of what it does.>
+    Copyright (C) 2013  PCMan <email>
+
+    This program is free software; you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation; either version 2 of the License, or
+    (at your option) any later version.
+
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License along
+    with this program; if not, write to the Free Software Foundation, Inc.,
+    51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA.
+*/
+
+#ifndef LXIMAGE_IMGURPROVIDER_H
+#define LXIMAGE_IMGURPROVIDER_H
+
+#include <QNetworkAccessManager>
+
+#include "provider.h"
+
+namespace LxImage {
+
+/**
+ * @brief Create uploads to Imgur's API
+ */
+class ImgurProvider : public Provider
+{
+    Q_OBJECT
+
+public:
+
+    virtual Upload *upload(QIODevice *device);
+
+private:
+
+    QNetworkAccessManager mManager;
+};
+
+}
+
+#endif // LXIMAGE_IMGURPROVIDER_H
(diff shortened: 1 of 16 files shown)


src/upload/imgurprovider.h
```diff
@@ -0,0 +1,47 @@
...
+{
+    Q_OBJECT
+
+public:
+
+    virtual Upload *upload(QIODevice *device);
+
+private:
+
+    QNetworkAccessManager mManager;
```
palinek: Just curious... I don't know how "heavy" the `QNetworkAccessManager` objects are, but the Qt doc says: `One QNetworkAccessManager should be enough for the whole Qt application.`.... then wouldn't it be feasible to make this object be a static member of the `Provider` class?

src/upload/imgurprovider.h
```diff
@@ -0,0 +1,47 @@
+/*
+    <one line to give the program's name and a brief idea of what it does.>
+    Copyright (C) 2013  PCMan <email>
```
palinek: Don't you want to gain the credits intentionally or is it just a copy/paste?

nathan-osman: I looked at the other headers and wondered the same thing :smile: I'd be more than happy to put my name on the new files if that's not a problem.

src/upload/uploaddialog.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "uploaddialog.h"
+
+using namespace LxImage;
+
+ImgurProvider gImgurProvider;
+
+UploadDialog::UploadDialog(QWidget *parent, const QString &filename)
+    : QDialog(parent),
+      mFile(filename),
+      mState(SelectProvider),
```
palinek: ...please, reorder the initialization:
```
lximage-qt/src/upload/uploaddialog.cpp:38:7: warning: field 'mFile' will be initialized after field 'mState' [-Wreorder]
      mFile(filename),
      ^
```

src/upload/uploaddialog.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    mState = UploadInProgress;
+    updateUi();
+}
+
+void UploadDialog::updateUi()
+{
+    // Show the appropriate control given the current state
+    ui.providerComboBox->setVisible(mState == SelectProvider);
+    ui.progressBar->setVisible(mState == UploadInProgress);
```
palinek: IMO the progress bar should remain visible (with the 100%) after the upload is completed... what do you think?

nathan-osman:
I believe the two additional commits addresses each of the issues.

src/upload/uploaddialog.cpp
nathan-osman: I suppose it could remain shown - it would appear above the line edit that shows the URL of the newly uploaded file. My personal preference is to have it disappear after the upload completes. Due to the way most image upload services work, the "100%" would be shown briefly for a few seconds while the image is being processed anyway.

src/upload/uploaddialog.ui
```diff
@@ -0,0 +1,75 @@
...
+    <widget class="QProgressBar" name="progressBar">
+     <property name="enabled">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_2">
+     <item>
+      <widget class="QLineEdit" name="linkLineEdit"/>
```
palinek: Wouldn't it be better to make the URL edit disabled (to not allow editing by hand)?

nathan-osman: Yes, but then the user can't select the text to copy it to the clipboard.

palinek: Heh... of course, I meant the `readOnly` attribute...

palinek:
GTM

agaida:
@palinek - any risks if rolling it out with 0.12.0?

palinek:
> any risks if rolling it out with 0.12.0?

Just a new highlight in release notes :-)

agaida:
would you be so kind and merge it  - please add the note, i'm busy with building the testpackages

nathan-osman:
Sorry, I had to add one more commit to fix a bug. Description is in the commit message.

agaida:
@nathan-osman - no problem, very appreciated

agaida:
Apply, compile, works for me™ - GTM

PR #117, opened 2017-10-21
Add ImageShack upload provider

nathan-osman:
This pull request adds an upload provider for ImageShack, using their public API.


agaida:
GTM - i will merge it as an experimental feature - the functionality is good, but need some final polish at all.

PR #120, opened 2017-12-05
cmake: Handle CMP0071

luis-pereira:
CMAKE_AUTOUIC marks the files it generates.
We mark the ones qt5_add_dbus_adaptor() generates.
Fix for lxde/lxqt#1420.


agaida:
GTM

PR #141, opened 2018-06-14
Fixes issue 109

### LINKED ISSUE #109: Can overwrite an existing image without prompt on saving
##### Expected Behavior
lximage-qt shouldn't overwrite any image without prompt while saving.

##### Current Behavior
It can overwrite without prompt when the chosen save name doesn't have any extension.

##### Possible Solution
Check for the existence of the image and show a prompt when needed.

##### Steps to Reproduce (for bugs)

1. Put an image with the name `X.png` in some folder.
2. Open another image with lximage-qt and save it with the name `X` *but without an extension* into that folder.
3. It'll be saved with the name `X.png` and will overwrite the previous image without prompt.

##### System Information

* Every LXQt component is the latest.

igordsm:
Add confirmation dialog when using "Save as" with a filename without extension. Fixes #109

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -278,12 +278,17 @@ QString MainWindow::saveFileName(QString defaultName) {
   }
   // FIXME: should we generate better filter strings? one format per item?
 
-  QString fileName = QFileDialog::getSaveFileName(
-    this, tr("Save File"), defaultName, tr("Image files (%1)").arg(filterStr));
+  QString fileName;
+  QFileDialog diag(this, tr("Save File"), defaultName, 
+               tr("Image files (%1)").arg(filterStr));
+  diag.setAcceptMode(QFileDialog::AcceptMode::AcceptSave);
+  diag.setFileMode(QFileDialog::FileMode::AnyFile);
+  diag.setDefaultSuffix("png");
+  
+  if (diag.exec()) {
+    fileName = diag.selectedFiles().at(0);
+  }
 
-  // use png format by default if the extension is not set
-  if(!fileName.isEmpty() && fileName.indexOf('.') == -1)
-    fileName += ".png";
   return fileName;
 }
 


tsujan:
You don't need a separate message box or even an existence check for this. Just don't use static functions but do something like:
```c++
...
QFileDialog dialog(this);
dialog.setAcceptMode(QFileDialog::AcceptSave);
// put appropriate lines of code here
if(dialog.exec()) {
    fileName = dialog.selectedFiles().at(0);
}
...
```

tsujan:
BTW, the condition `if(!fileName.isEmpty() && fileName.indexOf('.') == -1)`, which exists in the original code, isn't good; it doesn't cover cases like "THIS.NICE.IMAGE". However, I know that your goal wasn't fixing it.

jubalh:
And: https://gist.github.com/robertpainsi/b632364184e70900af4ab688decf6f53 :-)

igordsm:
@tsujan I might be missing your point, but the code you suggested has the same behavior as the current code. If I select an existing file the confirmation dialog appears as expected, but if I type a name manually _without extension_ the file still gets overwritten without confirmation. From my understanding, this happens because we add the extension *after* the dialog is dismissed, so `QFileDialog` cannot check whether the file exists or not because it does not know what extension we add. 

I am also coding a better detection for file names, so your second point will be addressed in the next PR udpate.

tsujan:
@igordsm Sorry that I wasn't more specific. By not using the static function I mean getting the dialog, finding its input field and connecting to its signal `QLineEdit::editingFinished` (hence the comment: `// put appropriate lines of code here`). Try the following code instead of the static function `QFileDialog::getSaveFileName`:
```c++
  QString fileName;
  QFileDialog dialog(this);
  dialog.setAcceptMode(QFileDialog::AcceptSave);
  dialog.setWindowTitle (tr ("Save File"));
  dialog.setFileMode (QFileDialog::AnyFile);
  dialog.setNameFilter (filterStr);
  dialog.selectFile (defaultName);
  dialog.setOption (QFileDialog::DontUseNativeDialog);
  // get the input and put ".png" in front of the name if there's no extension (still an incomplete solution)
  auto lineEdits = dialog.findChildren<QLineEdit *>();
  if(lineEdits.size() == 1) {
    QLineEdit* input = lineEdits.at(0);
    connect (input, &QLineEdit::editingFinished, input, [input, &dialog] {
      dialog.selectFile(input->text().indexOf('.') == -1 ? input->text() + ".png" : input->text());
    });
  }
  if(dialog.exec()) {
    fileName = dialog.selectedFiles().at(0);
  }
```
It works as expected by using only file dialog functionalities.

However, once I tried this solution (a few minutes ago), I realized that if I used the LXQt file dialog by removing the line:
```c++
  dialog.setOption (QFileDialog::DontUseNativeDialog);
```
then, the input field wouldn't be found!! Of course, we should NOT set `QFileDialog::DontUseNativeDialog` in lximage-qt. Why our LXQt file dialog has such a behavior is a mystery that should be solved before using a code like the above one.

Your code works but, since the file dialog has an overwrite prompt and shows a message, IMO, it's better to use that functionality in some way.

igordsm:
Thanks for the in-depth explanation. Indeed I WAS missing something. I have limited experience with Qt, so I never thought about this solution. 

I'll take a look at the QFileDialog code and try to find what's missing.

tsujan:
Even this simple code works and shows a prompt when there's no extension, without a need to find the input:
```c++
  QString fileName;
  QFileDialog dialog(this);
  dialog.setAcceptMode(QFileDialog::AcceptSave);
  dialog.setWindowTitle (tr ("Save File"));
  dialog.setFileMode (QFileDialog::AnyFile);
  dialog.setNameFilter (filterStr);
  dialog.selectFile (defaultName);
  dialog.setOption (QFileDialog::DontUseNativeDialog); // LXQt file dialog doesn't work
  dialog.setDefaultSuffix("png");
  if(dialog.exec()) {
    fileName = dialog.selectedFiles().at(0);
  }
```
I think LXQt file dialog has a problem.

tsujan:
I opened https://github.com/lxqt/lxqt/issues/1504

tsujan:
@igordsm This is my suggestion:

The default suffix bug will be fixed by this PR: https://github.com/lxqt/libfm-qt/pull/216. After that, a code like that of https://github.com/lxqt/lximage-qt/pull/141#issuecomment-397817321 will be enough and there will be no need to check for the extension anymore.

It's user responsibility not to use file names like `A.B`, unless `B` is a really an image extension. If the user puts `A` into the input field, `A.png` will be the result of the above-mentioned code and if `A.png` already exists, a prompt dialog will be show automatically (after the above-mentioned PR is merged).

Under KDE, it's up to KDE devs to fix the problem in their file dialog.

igordsm:
@tsujan Did just what you suggested. Now I'm waiting for [lxqt/libfm-qt#216](https://github.com/lxqt/libfm-qt/pull/216) to be accepted and than I'll test this again.

tsujan:
GTM.

@igordsm, thanks! It works fine and fixes the problem of overwriting -- I already have https://github.com/lxqt/libfm-qt/pull/216

BTW, your previous code was also good but this one is better in that it leaves prompting to libfm-qt.

PR #152, opened 2018-06-30
Use mouse forward/back button to change image

jakar:
I find myself wanting to use these buttons out of habit, especially when thumbnail view isn't open. With the wheel, sometimes I'll accidentally skip over an image or two.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -487,6 +487,20 @@ bool MainWindow::eventFilter(QObject* watched, QEvent* event) {
         }
         break;
       }
+      case QEvent::MouseButtonPress: {
+        QMouseEvent* mouseEvent = static_cast<QMouseEvent*>(event);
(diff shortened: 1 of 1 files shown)


jakar:
Please review this carefully. The mouse event handlers in `ImageView` don't check the button, so this change, as is, might affect those. One of the following might be necessary:

* Return `true` after calling `on_action{Next,Previous}_triggered` in `MainWindow::eventFilter`.
* Test `event->button()` in `ImageView::mousePressEvent` and so on.

However, it seems like this depends on the desired behavior of mouse cursor. So maybe neither is necessary.

tsujan:
I'm not sure but this may interfere with the new annotation capability. Someone having a mouse with forward/backward buttons should test it. Can annotations be drawn with those buttons and without this patch? If yes, forward/backward buttons should be excluded from annotations first.

PR #176, opened 2018-10-14
use XFIXES_LDFLAGS

BabylonAS:
use XFIXES_LDFLAGS so Clang linker can find the libXfixes library (FreeBSD)

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -91,7 +91,7 @@ target_link_libraries(lximage-qt
     ${QT_LIBRARIES}
     ${EXIF_LIBRARIES}
     ${X11_LIBRARIES}
-    ${XFIXES_LIBRARIES}
+    ${XFIXES_LDFLAGS}
 )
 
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -91,7 +91,7 @@ target_link_libraries(lximage-qt
     ${QT_LIBRARIES}
     ${EXIF_LIBRARIES}
     ${X11_LIBRARIES}
-    ${XFIXES_LIBRARIES}
+    ${XFIXES_LDFLAGS}
```
palinek: IMO everywhere (pkg-config's configurations) should be used \*_LDFLAGS instead of \*_LIBRARIES ...

BabylonAS: Huh. I haven’t encountered problems with other *_LIBRARIES inclusions (except POLKIT_AGENT_LIBRARIES).

agaida:
Works fine, GTM

PR #180, opened 2018-10-30
Add ability to annotate images

nathan-osman:
This pull request adds a toolbar which provides a number of tools for
annotating an image:

lximage-toolbar

These are especially handy when using lximage-qt for capturing screenshots:

lximage-demo

Although functional, there is probably room for improvement. I'd love to hear some feedback on my implementation and how I can make it better.


tsujan:
This seems very interesting. I'll review it as soon as I find time.

src/imageview.cpp
```diff
@@ -101,13 +105,71 @@ void ImageView::mouseDoubleClickEvent(QMouseEvent* event) {
...
+      // Draw the path
+      QPainterPath path;
+      path.addEllipse(circleRect);
+      painter.fillPath(path, Qt::red);
+      painter.drawPath(path);
+
+      // Draw the text
+      painter.setPen(Qt::white);
+      painter.drawText(textRect.bottomLeft(), text);
+
```
tsujan: The above line doesn't draw the text centered. Instead, it should be:
```c++
      painter.drawText(textRect, Qt::AlignCenter, text);
```
which will result in centered texts:

center

tsujan:
(1) The toolbar icons don't look good, especially with 16-px toolbar buttons:

toolbar

The icons:
  * should be nice enough;
  * should be clearly seen with small sizes (the current icons have big spaces around them); and
  * should look good with light and dark backgrounds alike.

(2) Without undo/redo, this feature won't be very practical.

(3) The annotations disappear when the image is rotated before saving.

tsujan:
@agaida A lot of work was still needed ;) Undo-redo isn't easy.

agaida:
Upps :)

tsujan:
> Upps :)

I'll try to add the missing parts.

Also, it would be great if @nathan-osman could make branches here.

nathan-osman:
This one wasn't quite ready yet. But I can help fix things up where possible. I'll try to tackle these during the coming week as time permits. Do you have any suggestions for improving the icons?

tsujan:
> Do you have any suggestions for improving the icons?

I can do that part because I have access to various Qt styles.

The most important thing about this feature is undo/redo, IMHO. The last time I implemented undo/redo from scratch was in a GTK+ app and in C, several years ago ;)

agaida:
sorry for being pain in the arse, friends :)

tsujan:
@agaida What should we do? There are serious problems:

 * Lack of undo/redo (undo/redo is necessary but hard to implement).
 * Disappearance on any transformation.
 * It doesn't work with PNG when the scale is 1.
 * It doesn't work with SVG at all.

Frankly, I can't fix them -- no incentive at all because the app is a viewer, not an editor.

agaida:
so remove the annotation things alltogether?

tsujan:
My suggestion is a little weird: Keep it and wait for someone to fix its issues at some point. My reason is that it's hidden from the user by default (in git) and doesn't interfere with other functionalities.

Is it acceptable to you?

agaida:
It isn't weird - it's the status quo since i merged this by mistake - so yes, it is acceptable.

tsujan:
Good. Who knows, I might do some things with it in future. If we receive more bug reports about it, we could say it's experimental -- which is the truth.

PR #181, opened 2018-10-31
Implement MRU list

nathan-osman:
This pull request adds a menu that shows recently opened files, enabling the user to quickly return to an image that was opened previously:

lximage-mru

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -23,6 +23,7 @@ set(lximage-qt_SRCS
     screenshotselectareagraphicsview.cpp
     settings.cpp
     graphicsscene.cpp
+    mrumenu.cpp
 
     upload/imageshackprovider.cpp
     upload/imageshackupload.cpp
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -47,6 +47,7 @@
 #include <libfm-qt/fileoperation.h>
 #include <libfm-qt/folderitemdelegate.h>
 
+#include "mrumenu.h"
 #include "upload/uploaddialog.h"
 
 using namespace LxImage;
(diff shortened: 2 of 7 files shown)


tsujan:
Thank you for taking time and making lximage-qt better! I hope you'll continue doing so.

This PR has the following problems:

 * The list shouldn't repeat a file. When a file is opened again, it should be moved to the top, not repeated.
 * The list only shows the file name. It's better to show the full path because a name can be in different paths (also (1) is related to this).
 * In all apps I've seen, the "Clear" item is at the end, after a separator, not at the top (for the user not to click it inadvertently or...).
 * The number of recent files should be configurable from 0 (don't remember recent files and don't show any menu) to 10 (or 5).

nathan-osman:
I moved most of the code to a separate class (`MruMenu`) to make things a bit more organized. I also did my best to address all but the last item on your list.

> The list shouldn't repeat a file. When a file is opened again, it should be moved to the top, not repeated.

This is the new behavior - if a file is added to the list, it is first compared to the files already in the list and if present, it is simple moved to the top.

> The list only shows the file name. It's better to show the full path because a name can be in different paths (also the previous issue is related to this).

Done.

> In all apps I've seen, the "Clear" item is at the end, after a separator, not at the top (for the user not to click it inadvertently or...).

Also done:

lximage-mru

Please let me know what you think about the new changes. If you are happy with them, I'll write the remaining code necessary for configuring the number of items shown in the menu.

tsujan:
@nathan-osman Thank you for the hard work!
I'm a little busy now; will read your code soon.

tsujan:
@nathan-osman I read your code and tested it. It's clean and efficient. Configuring the number of recent items can be done later, in another PR. So:

GTM.

There's just one minor thing that, IMO, isn't an issue: When the list is cleared, the Clear item isn't disabled until lximage-qt is restarted. However, since nothing bad will happen if it's clicked, I don't see it as a problem.

tsujan:
BTW, `mrumenu.cpp` is completely your work; so, you could add your name and email to the top comment.

agaida:
@nathan-osman  - would you be so kind and rebase?

tsujan:
@nathan-osman Also add your name on the top of `mrumenu.cpp` if you want.

nathan-osman:
I have rebased the pull request, added my name to the header comment, and fixed the issue with the "Clear" item being disabled.

agaida:
Cool, thanks a lot.

PR #199, opened 2018-11-10
Add reload button

nathan-osman:
This pull requests adds a button to the menu and toolbar for reloading the current image.


tsujan:
Thanks! GTM.

There is a bug in lximage-qt that I found only after testing this but it isn't related to the PR. I'll report it later.

agaida:
@nathan-osman thanks and nice that you are back

PR #214, opened 2018-12-18
Codestyle improvements

shlyakpavel:
Hello :)
It's me again with yet another pull request on code style. I really hope this has not become annoying yet and there is a sense in it...
I have also checked the project with PVS studio, but I couldn't find any serious problem with it. Great job!

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -238,7 +238,7 @@ void ImageView::zoomOriginal() {
   queueGenerateCache();
 }
 
-void ImageView::setImage(QImage image, bool show) {
+void ImageView::setImage(const QImage& image, bool show) {
   if(show && (gifMovie_ || isSVG)) { // a gif animation or SVG file was shown before
     scene_->clear();
     isSVG = false;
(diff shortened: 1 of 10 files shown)


tsujan:
GTM.

@shlyakpavel Please ignore C++11 raw string literals for now because there is this Qt bug about them: https://bugreports.qt.io/browse/QTBUG-42736 . It doesn't affect my apps (FeatherPad and FeatherNotes -- I'll make a commit for the latter soon) but I'm not sure about LXQt apps.

agaida:
@shlyakpavel  - no, your PRs are not annoying. they are highly appreciated :sunglasses: 

But again: please - keep them small. 

I know it is hard and is maybe not the most natural way to do things. In the current case a split into:
* PR for Null pointers
* PR Const reference
* PR Intendations
would be the better approach - a reviewer had only to review the same thing - that speed up things

Otherwise: GTM - thanks a lot :)

agaida:
@shlyakpavel - trust me, i'm not overly picky in this regard - my entry patches into LXQt were about 6.000 -10.000 lines each and really not easy to review fore mere humans - lesson learned, we should not do this again :D

shlyakpavel:
@agaida got you. I'll try to make smaller PRs next time :)
@tsujan thanks for the info about Qt bug. I didn't notice any regression here, but I'll be more careful with it.

tsujan:
> I didn't notice any regression here

@shlyakpavel 

Yes, it complies well but I'm worried about translation updates. A bad example is mentioned at that bug report. Hopefully, we won't encounter it here :)

By the way, I wanted to add a little explanation to what @agaida correctly said:

You may make a big PR about a single issue -- correcting references in function arguments *or* correcting `NULL`s *or* any other contribution. The size isn't important at all. What makes reviewing easier is that a PR is only about a single issue. That way, they will be merged faster too.

PR #220, opened 2019-01-07
ui: rename user-visible 'toolBar_2', and hide it in full screen mode

Chiitoo:
Seeing 'toolBar_2' looks just a little bit odd in the context menu.  :]

I first went with 'Toolbar 2', but then I realised its name is already 'annotationsToolBar' elsewhere, so I decided to go with that instead.

I'm not sure if the intention was to not hide this one in full screen mode, so as to be able to annotate while in said mode.  I can de-select the toolbar, but the setting isn't remembered (and if it was remembered, it would probably be hidden when exiting full screen mode too, which might be weird).

That said, it would be nice to be able to remove, and keep them both bars removed via the preferences, or perhaps the view settings.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1008,6 +1008,7 @@ void MainWindow::changeEvent(QEvent* event) {
       ui.view->setFrameStyle(QFrame::NoFrame);
       ui.view->setBackgroundBrush(QBrush(app->settings().fullScreenBgColor()));
       ui.toolBar->hide();
+      ui.annotationsToolBar->hide();
       ui.statusBar->hide();
       if(thumbnailsDock_)
         thumbnailsDock_->hide();
@@ -1034,6 +1035,7 @@ void MainWindow::changeEvent(QEvent* event) {
       }
       ui.menubar->show();
       ui.toolBar->show();
+      ui.annotationsToolBar->show();
       ui.statusBar->show();
       if(thumbnailsDock_)
         thumbnailsDock_->show();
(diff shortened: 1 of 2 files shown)


Chiitoo:
Ah, that PR explains quite a bit.  Had a feeling it wasn't quite done.  :]

Also, I forgot to check this before this PR, but is the string I changed used for the translations?

That is, should it also be changed in the 'ts' files?  For example 'src/translations/lximage-qt.ts':

```
169     <message>
170         <location filename="../mainwindow.ui" line="175"/>
171         <source>toolBar_2</source>
172         <translation type="unfinished"></translation>
173     </message>
```

Don't see greping around where else this 'toolBar_2' might be referring to.

I never quite remember how them translations tend to work.

yan12125:
LGTM, too. @tsujan how do you think about leaving refinement work of this toolbar to future PRs/commits?

tsujan:
>  @tsujan how do you think about leaving refinement work of this toolbar to future PRs/commits?

Yes. I have plans for it: (1) hide it by default (who needs so many buttons by default?); (2) use something like https://github.com/tsujan/FeatherPad/blob/master/featherpad/svgicons.cpp to have clear symbolic SVG icons instead of those ugly icons; (3) improve the code (add undo/redo).

The only thing I don't have is time. Good patches would be very appreciated.

yan12125:
> hide it by default (who needs so many buttons by default?)

That makes sense, but it will interfere with the "hide it in full screen mode" change in this pull request. @Chiitoo I think it's reasonable to remember whether toolbars are hidden or not in the settings. Do you have time to look into it in near future?

tsujan:
> but it will interfere with the "hide it in full screen mode" change in this pull request. 

Whoever will make the patch (I or someone else) will be responsible for that.

I agreed with a little image editing because it was handy, not because it was necessary. This is an image *viewer* after all and those buttons shouldn't be visible by default -- they're only a "bonus".

PR #233, opened 2019-03-08
Fix the file open string

agaida:
File -> File don't make sense and is not consistent with nearly all other™
Software around (Inkscape, Gimp, LO, Geany and so on)

Should be Open… (with Ellipsis instead of three point). We should change
'...' to '…' projectwide, all people will hate us and esp. me, esp the translators,
but hey - we really should do that.

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -201,7 +201,7 @@
      <normaloff>.</normaloff>.</iconset>
    </property>
    <property name="text">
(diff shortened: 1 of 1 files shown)


agaida:
Additional: Weblate has such a button for a reason :sunglasses:

tsujan:
Right!

tsujan:
Decided to work on lximage-qt.

PR #241, opened 2019-04-28
Fixed showing first image on loading a directory

tsujan:
Previously, no image was shown.

Also, select the current image on file opening and the current directory on directory opening if they exist; otherwise, open the home folder.


tsujan:
@agaida Please save these fix PRs from piling up ;) Features were added to lximage-qt but its issues weren't addressed.

Issue #196, opened 2018-11-03
An option to show image boundary


tsujan:
Like this (not a screenshot of lximage-qt):

![boundary](https://user-images.githubusercontent.com/981076/47956287-5dc4fc00-dfb7-11e8-9864-866e2fd85eb3.png)

Needed to check the space around small icons (for making icon themes, for example).

PR #252, opened 2019-05-27
Optionally show image outline

tsujan:
Closes https://github.com/lxqt/lximage-qt/issues/196

It is useful for checking the space around an image/icon.

Also found and fixed a small hidden bug about showing next/previous image.

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -40,13 +40,15 @@ ImageView::ImageView(QWidget* parent):
   QGraphicsView(parent),
   scene_(new GraphicsScene(this)),
   imageItem_(new QGraphicsRectItem()),
+  outlineItem_(new QGraphicsRectItem()),
   gifMovie_(nullptr),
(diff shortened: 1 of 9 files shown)


tsujan:
@agaida If you check this, the following properties should work fine:

 (1) The outline is like this (its toggling shortcut is `O`):

outline

 (2) It is rotated with the image.
 (3) It changes color when the background is changed from light to dark and conversely (for example, by full-screening) to have enough contrast with it.
 (4) It can be shown with startup (in Preferences).

tsujan:
News came that the whole planet waits for this PR to be decided (see https://github.com/lxqt/lximage-qt/pull/254#issuecomment-497940090).

tsujan:
An important bug (in HDPI) needs to be fixed and its fix needs this. So, I have to tell a lie:

Since the PR is straightforward, I merge it and accept the responsibility. (The last part wasn't a lie).

Issue #242, opened 2019-05-08
Images are blurry with scale 1.5


bosim:
Images seem blurry when running with scaling, e.g.

    QT_FONT_DPI=96
    QT_SCALE_FACTOR=1.5

Maybe lximage-qt needs same fix as gwenview:
https://github.com/KDE/gwenview/commit/97e7de52276b46bc99b64117abfee0965e6f58e6

##### Expected Behavior

Crystal clear images

##### Current Behavior

Blurry images

##### Possible Solution

See above

##### Steps to Reproduce (for bugs)

1. QT_SCALE_FACTOR=1.5
2. lximage-qt
3. open an image and observe it is blurry

##### System Information

* Distribution & Version: Debian buster
* Kernel: 4.19
* Qt Version: 5.11
* libfm-qt Version: 
* libqtxdg Version: 
* lxqt-build-tools Version: 
* Package version:

tsujan:
Good point!

But w have a bigger issue: https://github.com/lxqt/libqtxdg/issues/178

tsujan:
I self-assigned because I fixed the problem here — waiting for https://github.com/lxqt/lximage-qt/pull/252 to be decided because of conflicting code lines.

tsujan:
@bosim This branch has the fix: https://github.com/lxqt/lximage-qt/tree/hdpi

The PR is here: https://github.com/lxqt/lximage-qt/pull/257

PR #254, opened 2019-05-31
Fix for blurry rendering of images when scale factor is fractional

bosim:
Fixes issue #242


tsujan:
The problem is deeper than scaling the cached image (you could see it with gif animations too — they aren't cached) and increasing the scale factor doesn't solve it.

Please wait until https://github.com/lxqt/lximage-qt/pull/252 is merged (because of conflicts) and then we'll discuss it

tsujan:
@agaida Please don't merge it -- it'll be superseded by a working PR.

bosim:
> @agaida Please don't merge it -- it'll be superseded by a working PR.

I will close if you are working on a solution that will cover it all.

tsujan:
> I will close if you are working on a solution that will cover it all.

The solution is ready but should wait a little because of PR conflicts.

PR #259, opened 2019-07-28
added extra option -d

alexzk1:
Well, sometimes I want screenshot fast-running things, like flying inside game or so. I don't have time to click all that boxes and/or it may crash game.
So this one just addsoption -d <folder> where program just dumps files with robust namings.
Also some needed changes to avoid copy-paste.


tsujan:
Thanks for your contribution! Please:

(1) Clean up the PR (`CMakeLists.txt.user`?!).

(2) Respect the code style of the original author (number of spaces, no space between `if` and `{`, etc. -- a big part of the diff is about that).

alexzk1:
well, yeh, I changed code styling on late commits ..however I still guessing options for it, can you provide styling options for "Artistic Style" ? 
Also different files has different settings.

tsujan:
@alexzk1 Thanks! It's much better now.

Please be patient as its review might take a while -- unless another dev already has the required time.

tsujan:
One more thing: This part of lximage-qt should be completely rewritten, based on KWindowSystem. We should remove X11 codes as far as possible. But that's a job for future.

src/screenshotdialog.cpp
```diff
@@ -183,9 +210,84 @@ void ScreenshotDialog::doScreenshot() {
...
 
   deleteLater(); // destroy ourself
 }
+
+static QString buildNumericFnPart() {
+  //we may have many copies running with no gui, for example user presses hot keys fast
+  //so they must have different file names to save, lets do it time + pid
+  const auto now = QDateTime::currentDateTime().toMSecsSinceEpoch();
+  const auto pid = getpid();
+  return QString(QString::fromLatin1("%1_%2")).arg(now).arg(pid);
```
tsujan: Please change this to:

```c++
return QStringLiteral("%1_%2").arg(now).arg(pid);
```

src/screenshotdialog.cpp
```diff
@@ -183,9 +210,84 @@ void ScreenshotDialog::doScreenshot() {
...
+  return (result.isEmpty())?QString::fromUtf8("UKNOWN"):result;
+}
+
+void ScreenshotDialog::cmdTopShotToDir(const QString &path) {
+
+  WId activeWid = activeWindowId();
+  const QRect rect = (activeWid) ? windowFrame(activeWid) : QRect{0, 0, -1, -1};
+  QImage img{takeScreenshot(QApplication::desktop()->winId(), rect, false)};
+
+  const QString filename = path % QString::fromLatin1("/") % getWindowName(activeWid) % QString::fromUtf8("_") % buildNumericFnPart();
```
tsujan: Please change all instances of `QString::fromLatin1()` to `QStringLiteral()`.

src/screenshotdialog.cpp
```diff
@@ -183,9 +210,84 @@ void ScreenshotDialog::doScreenshot() {
...
+            }
+            result = QString::fromStdString(ss.str());
+            break;
+          }
+        }
+
+      }
+    }
+  }
+  return (result.isEmpty())?QString::fromUtf8("UKNOWN"):result;
```
tsujan: Please change all instances of `QString::fromUtf8()` to `QStringLiteral()`.

alexzk1:
Done. Changed string join too.
However can't compile now because reverted system libfm-qt to older. Older does not compile, newer which compiles, leads to crashes of file manager. 
So try yourself too to see if I didn't do typos.

tsujan:
> newer which compiles, leads to crashes of file manager.

LXQt components should be upgraded together. For example, when libfm-qt is upgraded, pcmanfm-qt, lximage-qt and lxqt-qrchiver should be too.

> So try yourself too to see if I didn't do typos.

I will. The code is good, in general.

Why did you add PID to file name? Wasn't the current date in milliseconds enough?

alexzk1:
>Why did you add PID to file name? Wasn't the current date in milliseconds enough?

Because u can launch couple copies very fast (on many cores it is easy) and all them may have same ms stamp. So PID ensures name is unique, and still I do extra check there and it may add extra sequental number. For example that may happen if between 2 screenshots network time correction happened.

tsujan:
OK. I guess you've considered all possible scenarios.

alexzk1:
>OK. I guess you've considered all possible scenarios.
Not really, but I feel that probability to overwrite is very low.
To catch really ALL need atomic write :D That will require to have some running daemon which manages names...

tsujan:
I see no problem. GTM.

Would you please squash all of your commits, so that I can merge it?

tsujan:
Never mind! GitHub had a button for that :)

tsujan:
Oh, GitHub's squash-and-merge button wasn't good :( I shouldn't use it anymore.

alexzk1:
Offtopic:
That's why I needed that -d option, to do that: 
image

tsujan:
> That's why I needed that -d option, to do that:

I knew. It was a good addition.

However, we really need to get rid of X11 codes in lximage-qt, as far as possible.

PR #260, opened 2019-07-30
updated to check if directory exists/created, otherwise fallback to $…

alexzk1:
…HOME


tsujan:
Please choose more concise, legible and independent titles for your future PRs! You could imagine them being included in a changelog.

Issue #283, opened 2019-10-29
Minimal GUI


glpzzz:
This is something I did for my self perhaps is interesting for everyone. I like image viewers with minimal GUI, as sxiv, so I did this changes to my local lximage-qt

##### Expected Behavior
Ability to hide all toolbars, menubar and status bar,  and the program remembers this behavior between executions.

##### Current Behavior
The annotations toolbar can be hidden, the main toolbar too but not from view menu, just with right click in toolbars area. Also nor the menu nor the status bar can be hidden. Just in fullscreen status

##### Possible Solution
I replicated what is done with the annotations toolbar for the others

##### Context
This is just taste thing. I like this way and perhaps other users too. It can be done in fullscreen mode but fullscreen is not working if you can keep the other windows in the same view.

tsujan:
There's no need to add a menu-item for it. Please add a "Compact mode" to Preferences instead and make a pull request (PR)! In that way, your goal will be achieved without cluttering the GUI.

glpzzz:
And there is a menu item for annotations. And in other lxqt apps like lxqt-archiver there is a toggle  visibility option for every bar in the view menu. 

But I like the compact mode idea

glpzzz:
Also added Tab as shortcut por compact interface mode.

tsujan:
Done in https://github.com/lxqt/lximage-qt/commit/e3a4788d82e541f8e6e57a40342db3e5741e8d52, https://github.com/lxqt/lximage-qt/commit/7d5ba9c910d5038251e1b26d5ff3cb4691f03b7d and some other commits.

stefonarch:
Give me a call when all is over, to update translations ;)

tsujan:
Sure.

tsujan:
@stefonarch 
It isn't finished yet but I don't think it'll be finished soon. You could update translations if you think an update is needed, knowing that it should be followed by another update later.

PR #284, opened 2019-10-30
WIP: Glpzzz #283 minimal gui

glpzzz:


glpzzz:
Yes! I add this today! 

On October 30, 2019 5:36:05 AM GMT-04:00, tsujan <notifications@github.com> wrote:
>tsujan requested changes on this pull request.
>
>This will be a nice feature. I like the idea and your implementation is
>good but it has two problems. Please fix them:
>
>(1) The shortcuts aren't accessible in the compact mode. You could add
>something like this to `on_actionCompact_triggered` before
>showing/hiding widgets:
>
>```c++
>  if(checked){
>    addActions(ui.menubar->actions());
>  }
>  else {
>    const auto actions = ui.menubar->actions();
>    for(QAction* action : actions) {
>      removeAction(action);
>    }
>  }
>```
>(2) Please never use `Tab` as shortcut unless you have handled it in
>the code. The reason is that `Tab` can change focus and, if it's a
>shortcut too, compilations and bugs will be imminent.
>
>`C` will be a good shortcut for the compact mode. (We'll add a shortcut
>editor later.)
>
>Also, please add the shortcut `Ctrl+F` to `actionZoomFit` for it be
>usable in the compact mode.
>
>
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/lxqt/lximage-qt/pull/284#pullrequestreview-309064225

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

tsujan:
> Yes! I add this today!

Take your time.

After you make the changes and this PR is merged, I should fix a small issue elsewhere in the main code (it's related to the PR but isn't caused by it).

tsujan:
Oh, there's a third problem that I forgot to mention: The compact mode is broken on un-full-screening.

The problem starts here:

https://github.com/lxqt/lximage-qt/blob/71cdcebb3456160180de8abec11cfeb19f8eecd0/src/mainwindow.cpp#L1175

The solution is something like this (replacements and additions):

```c++
      if(!ui.actionCompact->isChecked()) {
        // now we're going to re-enable the menu, so remove the actions previously added.
        const auto actions = ui.menubar->actions();
        for(QAction* action : actions) {
          removeAction(action);
        }
        ui.menubar->show();
        ui.toolBar->show();
        ui.statusBar->show();
      }
```

Please note that I removed the condition `if(!action->shortcut().isEmpty())` here; it was a mistake (a small bug).

The annotation toolbar and thumbnail dock s already dealt with correctly.

tsujan:
And the fourth problem: The compact mode shouldn't be toggled when the window is in the full-screen state. Please fix it too.

src/mainwindow.cpp
```diff
@@ -1156,37 +1167,37 @@ void MainWindow::changeEvent(QEvent* event) {
...
+      
+      if(!ui.actionCompact->isChecked()){
+        ui.menubar->show();
+        ui.toolBar->show();
+        if(ui.actionAnnotations->isChecked()){
           ui.annotationsToolBar->show();
+        }
+        ui.statusBar->show();
+        if(thumbnailsDock_){
+          thumbnailsDock_->show();
```
tsujan: Thumbnail dock and annotation toolbar should be *outside* the block, as mentioned at https://github.com/lxqt/lximage-qt/pull/284#issuecomment-547873735

Also, there's no need to the method `toggleActions()`. Its name damages code readability too.

Also, the existence of shortcuts shouldn't be checked on adding/removing actions (→  https://github.com/lxqt/lximage-qt/pull/284#issuecomment-547873735)

glpzzz: Ok @tsujan. I will review this again. I created the function to avoid rewriting the same code for the case of fullscreen and compact. I think I'll more explanation on what we want to achieve. For me is working good now. Can you explain me again please?

tsujan: > For me is working good now.

That's the problem: "For me". A code should work under all circumstances, not only when the app is used in a specific way.

> Can you explain me again please?

Please read my above explanations again. I've already explained some things twice.

tsujan:
Done.

Issue #263, opened 2019-08-14
Problem with printing images from lximage-qt


travis-82:
Sorry devs, It might seems ridiculous, but have you ever tried to print images from lximage-qt?

On Manjaro (kernel 4.19 lts, Lxqt 0.14.1) I can not print images correctly using my hp 4300 printer. Image printed as a quarter of full size per sheet. No matter what option I choose, it needs 4 sheet to print the image. I set Hp Device Manager option for image printing to  "Fit to page", but lximage keep to print image quarterly. This problem doesn't happen with other image viewers like GIMP and Nomacs. I didn't check this problem on other distros as Manjro is the only distro which I use.

Sorry for my english

tsujan:
lximage-qt doesn't print anything; CUPS does (Hello, Apple!). All options are in Qt print dialog.

gerhardbeck:
Hi,
I wanted to join @travis-82 
I tried to print a picture on a 10x15 picture paper. I configured all right in the QT print dialog, but in LXimage is one important dialog missing: "resize to paper size".
You find it e.g. in qpdfview
Can it be that there are different ways in different programs which settings are shown in the print dialogue?

tsujan:
My bad; I shouldn't close this.

Reopened to be investigated. Printing to PDF isn't good and now I think the problem is in our code.

tsujan:
OK, a glance at the code shows that the current behavior is intentional: the image is divided to "columns" when it can't be put into a single page.

However, IMO, that isn't practical. Instead, it's better to resize the image when it's too big. Other image viewers/editors I tested did it that way. We might want to add an option for it later but, for now, resizing seems better.

tsujan:
If you want to test it, the fix is here: https://github.com/lxqt/lximage-qt/pull/291

gerhardbeck:
Thanks @tsujan for having a second look at it!

tsujan:
You're welcome. It was a valid bug report; I should have been very busy with other codes when I closed it.

PR #291, opened 2019-11-26
Fixed image printing

tsujan:
Previously, a large image was divided and cut. Now, it fits into the page.

The code of the old behavior is commented out, not removed.

NOTE: I tested it only with PDF.

Fixes https://github.com/lxqt/lximage-qt/issues/263


gerhardbeck:
great!

PR #295, opened 2019-12-14
Remove deprecated QDesktopWidget

luis-pereira:
Maintaining functionality.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -34,10 +34,10 @@
 #include <QWheelEvent>
 #include <QMouseEvent>
 #include <QTimer>
+#include <QScreen>
 #include <QShortcut>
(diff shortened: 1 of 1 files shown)


src/mainwindow.cpp
```diff
@@ -34,6 +34,7 @@
 #include <QWheelEvent>
 #include <QMouseEvent>
 #include <QTimer>
+#include <QScreen>
```
tsujan: Would you please also remove `#include <QDesktopWidget>` (a few lines below)?

luis-pereira: sure. completely forgot it.

luis-pereira:
@tsujan Done. Thanks.

tsujan:
Thanks. GTM.

PR #298, opened 2020-01-10
screenshotdialog: use shorter suffix in the delay spinbox

Fat-Zer:
Use "sec" suffix instead of "seconds" for the delay spinbox, because
the second one doesn't play well with value==1 and looks even worse in
translations.


tsujan:
> ...because the second one doesn't play well with value==1

That could have been fixed but "sec" seems better.

Issue #301, opened 2020-02-12
Drop Maximization on open gif


TrueCat17:
On open gif-file (not animated too) viewer's window drops param "to maximazed"  
After all images shown with them real size, until window will be maximazed manually

(if open image in other format and press Left/Right to switch on gif - all ok)

* Distribution & Version: Debian 11 (Bullseye, unstable, Sid)
* Kernel: 5.2
* Qt Version: 5.12.5
* libfm-qt Version: libfm-qt6 ver. 0.14.1-12
* lxqt-build-tools Version: in repo 0.6.0-2

tsujan:
I didn't understand the English -- not a single sentence. Please be more specific by describing the issue in a step-by-step and reproducible way.

Please note: If you filled in the issue template, your report might be understandable.

TrueCat17:
Sorry, but I think, I make not bad description

1. Open, for example, PNG-image
2. Maximaze window (not resize, not fullscreen, button to maximazed near with button to close window: [_][O][X] - second)
3. Close viewer
4. Open prev image again - window of viewer will be maximazed
5. Close again
6. Open GIF-image
7. Window not maximazed - bug (!)

tsujan:
> Sorry, but I think, I make not bad description

You did but your second description is quite good. Please try to make your future reports like it. Thanks in advance.

It shows there is a discrepancy in opening gif/SVG images, on the one hand, and PNG/JPG ones, on the other hand.

PR #306, opened 2020-03-18
Fixed maximization at startup with SVG and GIF images

tsujan:
The root of this minor problem was deeper than meets the eye: with SVG and GIF, the maximization info was lost when we resized the window before showing it. This may be a bug in Qt. The patch remembers the maximization info and applies it when the window is going to be shown.

Fixes https://github.com/lxqt/lximage-qt/issues/301

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -62,6 +62,7 @@ MainWindow::MainWindow():
   image_(),
   // currentFileInfo_(nullptr),
   imageModified_(false),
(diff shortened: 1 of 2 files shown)


tsujan:
Until Alf comes back, I'll merge my tested libfm-qt based PRs if other devs don't review them for 3 weeks -- otherwise, PRs could overlap with each other. Of course, I take full responsibility for probable mistakes (and fix them).

PR #316, opened 2020-04-27
Added copy Path action

tsujan:
The user can assign a shortcut to it in the shortcut editor.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -909,6 +909,13 @@ void MainWindow::on_actionCopy_triggered() {
   clipboard->setImage(copiedImage);
 }
 
+void MainWindow::on_actionCopyPath_triggered() {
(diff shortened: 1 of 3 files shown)


selairi:
GTM.

Maybe, I would change the Edit menu from:
`-------------`
`  * Copy to clipboard`
`  * Paste from clipboard`
`-------------`
`  * Copy Path`  

to:
`-------------`
`  * Copy Image to Clipboard`
`  * Copy Image Path to Clipboard`  
`  * Paste from Clipboard`
`-------------`

Thanks for your work.

tsujan:
> Maybe, I would change the Edit menu from...

That's also a valid arrangement. I didn't want to mix `Copy Path` with real image copying/pasting: one is about a simple text, while the others are about images.

Issue #307, opened 2020-04-13
Drawing tools don't work


jimgreen2013:
##### Expected Behavior
draw a rectange or a circle

##### Current Behavior
can't draw a circle, nothing happened

##### Possible Solution
i check the system log with journalctl, there is some log like:
```bash
Apr 13 09:58:55 miPad lximage-qt[12057]: QCoreApplication::postEvent: Unexpected null receiver
```
Maybe there is a null pointer

##### System Information
* Distribution & Version: Fedora 31
* Kernel: 5.4.15
* Qt Version: 5.13.2
* libfm-qt Version: 
* libqtxdg Version: 
* lxqt-build-tools Version: 
* Package version: 0.14.1

Chiitoo:
Had a user report this a while ago, and I noticed things are actually drawn, sort of, just not shown.

If you save the file, they appear in the saved image.

tsujan:
As I mentioned elsewhere (sorry, don't have time to find where), the drawing tools were merged by mistake. IMO, lximage-qt shouldn't do image editing. I wanted to remove them but some users told me not to. So, I disabled them by default in git (the user can re-enable them).

In short, *the drawing tools have several bugs* and, until now, no one has had time to fix them.

jimgreen2013:
@Chiitoo thanks for your reply.
@tsujan totally understand what you're talking about, besides i am big fan of [Unix philosophy](https://en.wikipedia.org/wiki/Unix_philosophy)
> do one thing and do it better

however, consider a situation when you want a drawing tools, without an image viewer, you can't even find the right place to draw. So in my opinion, you shouldn't remove it.
And drawing tools is really very helpful in my work when i communicate with my colleagues via IM app.
Maybe i can try to look into the code and find what cause the bug, but not confident, because i'm not familiar lximage or lxqt code

tsujan:
> So in my opinion, you shouldn't remove it.

I won't.

I've found serious problems in its code but haven't found time to fix them, being busy with other codes. Its coder was working on it and I was reviewing his work it but we merged it by mistake. Since then, he hasn't touched it.

> Maybe i can try to look into the code...

Yes, please!

jimgreen2013:
@tsujan I think i have found the reason for this bug, according to the qt documentation:
> Warning: When the paintdevice is a widget, QPainter can only be used inside a paintEvent() function or in a function called by paintEvent().

[...]

PR #319, opened 2020-05-11
fix bug #307 : drawing rectangle doesn't show

jimgreen2013:
The reason for this bug is that when paintdevice is a widget,
QPainter can only be used inside a paintEvent() function or in a
function called by paintEvent()

this commit fix this bug by drawing the same rectangle in the scene

LXImage use QGraphicsView to display image and QGraphicsView is a
part of the Grapics View Framework. Usually by calling update(), a
widget is able to trigger paintEvent(). However, as far as i know,
in the Graphics View Framework, it's impossible for QGraphicsView
to tigger the paintEvent() by calling update()

so i choose the Graphics View Framwork way to draw the rectangle
something like: scene.addRect(...)

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -137,10 +137,14 @@ void ImageView::mouseReleaseEvent(QMouseEvent* event) {
       drawArrow(painter, startPoint, endPoint, M_PI / 8, 25);
       break;
     case ToolRectangle:
+      // Draw the rectangle in the image and scene at the same time
       painter.drawRect(QRect(startPoint, endPoint));
+      scene_->addRect(QRect(startPoint, endPoint), painter.pen());
       break;
     case ToolCircle:
+      // Draw the circle in the image and scene at the same time
       painter.drawEllipse(QRect(startPoint, endPoint));
+      scene_->addEllipse(QRect(startPoint, endPoint), painter.pen());
       break;
     case ToolNumber:
     {
@@ -161,15 +165,21 @@ void ImageView::mouseReleaseEvent(QMouseEvent* event) {
                         textRect.top() + (textRect.height() / 2 - radius),
                         radius * 2, radius * 2);
 
-      // Draw the path
+      // Draw the path in the image
       QPainterPath path;
       path.addEllipse(circleRect);
       painter.fillPath(path, Qt::red);
       painter.drawPath(path);
+      // Draw the path in the sence
+      scene_->addPath(path, painter.pen(), QBrush(Qt::red));
 
-      // Draw the text
+      // Draw the text in the image
       painter.setPen(Qt::white);
       painter.drawText(textRect, Qt::AlignCenter, text);
+      // Draw the text in the sence
+      QGraphicsTextItem *textItem = scene_->addText(text, painter.font());
+      textItem->setPos(textRect.x() - textRect.width() / 8 , textRect.y() - textRect.height() / 8);
+      textItem->setDefaultTextColor(Qt::white);
 
       break;
     }
@@ -573,8 +583,10 @@ void ImageView::drawArrow(QPainter &painter,
                           qreal tipAngle,
                           int tipLen) const
 {
-  // Draw the line
+  // Draw the line in the inmage
   painter.drawLine(start, end);
+  // Draw the line in the scene
+  scene_->addLine(QLine(start, end), painter.pen());
 
   // Calculate the angle of the line
   QPoint delta = end - start;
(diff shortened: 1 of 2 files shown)


tsujan:
Can be a good approach but please fix the following problem in it:

The annotations persist. They can be seen on other images especially without zooming but they're also visible with zooming, although for a moment. They aren't cleared by reloading either.

jimgreen2013:
@tsujan you're right, i ignore those situations.

now all annotations has been recorded in a list, the items will be removed when `setImage()` is called, 
so rotating the image will also result in removing the annotations, i'm not sure whether it's appropriate.

tsujan:
> so rotating the image will also result in removing the annotations, i'm not sure whether it's appropriate.

It's appropriate, IMHO.

Good to see you remove obstacles one by one. I'll test and review your code soon.

src/imageview.cpp
```diff
@@ -307,8 +318,11 @@ void ImageView::setImage(const QImage& image, bool show) {
     zoomFit();
   queueGenerateCache();
 
-  // clear the numbering
+  // clear the numbering and annotation items in the scene
   nextNumber = 1;
+  for (QGraphicsItem *annotation : annotations){
+      scene_->removeItem(annotation);
+  }
```
tsujan, line 325: There are three problems here:

 1. This can cause crash with SVG and GIF because `scene_->clear();` is used for them near the beginning of the current function, which makes each annotation a dangling pointer.
 2. There is a memory leak because `QGraphicsScene::removeItem()` doesn't delete the item.
 3. `annotations` should be cleared here.

So, I suggest you add `annotations.clear();` after `scene_->clear();` (see above) and use these lines here:
```c++
  // clear the numbering and annotation items in the scene
  nextNumber = 1;
  if(!annotations.isEmpty()) {
    for(QGraphicsItem *annotation : qAsConst(annotations)) {
        scene_->removeItem(annotation);
    }
    qDeleteAll(annotations.begin(), annotations.end());
    annotations.clear();
  }
```

jimgreen2013: @tsujan you're right, it will cause crash and memory leak
 I have fixed it. The new code is almost like the code you suggest, but with a minor adjustment. 
because we can also annotate gif image, the annotations list may also not be empty when it's gif image, so i put the code to the beginning of the function. as the commit log describes:

>  In setImage() function, when the image is gif or svg, scene_->clear()
    is called, which clear and delete the annotation items, so we just
    need clear the annotations list, delete or remove the annotations
    will make a crash
    on the other hand, for a normal image, scene_->removeItem() just
    remove the item but not delete it, so we need delete it by ourself
    to prevent a memory leak
    At last, we need clear the annotations list for both situations

tsujan: > At last, we need clear the annotations list for both situations

That's what I suggested above.

So far so good. I didn't find another problem by a glance at the PR. Will read its details soon.

src/imageview.cpp
```diff
@@ -161,15 +165,22 @@ void ImageView::mouseReleaseEvent(QMouseEvent* event) {
...
+      // Draw the path in the sence
+      annotations.append(scene_->addPath(path, painter.pen(), QBrush(Qt::red)));
 
-      // Draw the text
+      // Draw the text in the image
       painter.setPen(Qt::white);
       painter.drawText(textRect, Qt::AlignCenter, text);
+      // Draw the text in the sence
+      QGraphicsTextItem *textItem = scene_->addText(text, painter.font());
+      textItem->setPos(textRect.x() - textRect.width() / 8 , textRect.y() - textRect.height() / 8);
```
tsujan, line 181: Please avoid magical numbers! 8 is an arbitrary integer here.

The reason why the text isn't centered is that `QGraphicsTextItem` supports HTML text and Qt has an old issue when dealing with the position of HTML texts, incorrectly considering their HTML tags when positioning them.

Use `QGraphicsSimpleTextItem` instead. Then, these 5 lines (from the comment to `annotations.append(textItem);`) can be rewritten as:
```c++
      // Draw the text in the sence
      QGraphicsSimpleTextItem *textItem = new QGraphicsSimpleTextItem(text);
      textItem->setFont(font);
      textItem->setBrush(Qt::white);
      textItem->setPos(textRect.topLeft());
      scene_->addItem(textItem);
      annotations.append(textItem);
```

src/imageview.cpp
```diff
@@ -571,10 +599,13 @@ void ImageView::drawArrow(QPainter &painter,
                           const QPoint &start,
                           const QPoint &end,
                           qreal tipAngle,
-                          int tipLen) const
+                          int tipLen,
+                          QList<QGraphicsItem *> *annotations) const
 {
```
tsujan, line 604: Let's hope this is the last change before merging -- but no promise ;)

The extra variable `QList<QGraphicsItem *> *annotations` seems a little weird and isn't needed here; `annotations` is already a private variable. Instead, you could just drop `const` and use `annotations` directly.

jimgreen2013: the const has been removed

the progress of making the code better and better is a kind of pleasure~ feel free to point out anything that need to be improved~

jimgreen2013:
WOW, that's so cool!! Iit's my honour and so appreciated for your help ^_^

Issue #336, opened 2020-10-01
PDF files (and, probably, some other mimetypes) aren't listed in thumbnail view


tsujan:
Although PDF is supported, it doesn't show up in the thumbnail view. That creates problems like the following one:

##### Steps to Reproduce:

1. Put a PDF file that has image (most do) inside a folder that contains PNG/JPEG images.
2. Drag and drop the PDF into lximage-qt. It'll be shown.
3. Try to change the image by pressing the relevant toolbar buttons.

The image couldn't be changed.

##### System Information

Latest git LXQt.

PR #337, opened 2020-10-02
Consult Qt for supported mimetypes

tsujan:
Previously, the code used `Fm::FileInfo::isImage()`, which only checks if the mimetype starts with "image/", while Qt may support more.

Closes https://github.com/lxqt/lximage-qt/issues/336


luis-pereira (approved):
GTM.

PR #345, opened 2020-10-19
Start app on screen with mouse cursor when opening image

tsujan:
And resize it accordingly.


tsujan:
This is like https://github.com/lxqt/qterminal/pull/716. Merging...

PR #376, opened 2020-12-24
Populate MimeType property of lximage-qt.desktop automatically

alexhenrie:
This ensures that support for image/webp is advertised when available.

### DIFF
--- a/data/lximage-qt.desktop.in
+++ b/data/lximage-qt.desktop.in
@@ -5,4 +5,3 @@ Exec=lximage-qt %F
 StartupNotify=true
 Terminal=false
 Categories=Graphics;Viewer;RasterGraphics;2DGraphics;Photography;
-MimeType=image/bmp;image/gif;image/jpeg;image/jpg;image/png;image/tiff;image/x-bmp;image/x-pcx;image/x-tga;image/x-portable-pixmap;image/x-portable-bitmap;image/x-targa;image/x-portable-greymap;application/pcx;image/svg+xml;image/svg-xml;
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -78,7 +78,6 @@ include(LXQtTranslateDesktop)
 
 file(GLOB desktop_files_in ../data/*.desktop.in)
 lxqt_translate_desktop(desktop_files SOURCES ${desktop_files_in} USE_YAML)
-install(FILES ${desktop_files} DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/applications")
 
 add_executable(lximage-qt
     ${lximage-qt_SRCS}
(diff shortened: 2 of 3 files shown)


tsujan (commented):
There's no guarantee that `QImageReader::supportedMimeTypes()` gives the same list in all systems. So, a reasonable list is needed in the desktop file.

alexhenrie:
@tsujan That's the point--If the Qt image format plugins were installed when lximage-qt was compiled, then it's reasonable to assume that lximage-qt will be able to open `.webp` files after installation, and support for formats like `image/tiff` and `image/webp` should be advertised. If the plugins were not installed at compile time, then that's not a reasonable assumption and support for `image/tiff` and `image/webp` should _not_ be advertised. Currently the `.desktop` file always advertises support for `image/tiff` even if lximage-qt can't open `.tiff` files, and never advertises support for `image/webp` even if lximage-qt can open `.webp` files, which leads to a poor user experience.

Being able to get the correct list from `lximage-qt --supportedmimetypes` at runtime also has the advantage that the `.desktop` file could be updated automatically by a script if necessary.

Another solution would be to say that all of the image format plugins are required dependencies of lximage-qt and then list every format in the `.desktop` file. Or we could have some kind of compilation switch that removes the extra formats from the list if the packager says that they won't be supported. If either one of those solutions is what you want then let me know and I'll try to make it happen.

tsujan:
> If the plugins were not installed at compile time,...

Whose compile time? lximage-qt is packaged by distro maintainers, like any other app.

> Another solution would be to say that all of the image format plugins are required dependencies of lximage-qt

They simply aren't.

It seems you've forgotten mime type association, which exists in all DEs. In LXQt, its GUI is File Associations.

alexhenrie:
@tsujan I know that I can use the File Associations program, but what should the default behavior be? Do you think lximage-qt should adverise support for both `image/tiff` and `image/webp` by default even if they aren't likely to work?

tsujan:
> but what should the default behavior be?

Both FeatherPad and Kate can be used for coding in tens of programming languages. Consider the `MimeType` sections of their desktop files as a hint.

I'm afraid there's no problem to solve.

alexhenrie:
From [FeatherPad's .desktop file](https://github.com/tsujan/FeatherPad/blob/master/featherpad/data/featherpad.desktop): `MimeType=text/plain;`

From [Kate's .desktop file](https://invent.kde.org/utilities/kate/-/blob/master/kate/data/org.kde.kate.desktop): `MimeType=text/plain;`

This works because `text/x-c`, `text/x-c++`, `text/x-script.python`, etc. are all subclasses of `text/plain` (see the [shared-mime-info](https://gitlab.freedesktop.org/xdg/shared-mime-info/-/blob/master/data/freedesktop.org.xml.in) package). FeatherPad and Kate can open `text/plain` files, so they automatically also support all of those other types too.

Perhaps you've never run into this problem, but lximage-qt ***fails*** to open TIFF and WEBP files unless the [qt5-image-format-plugins](https://packages.ubuntu.com/xenial/amd64/qt5-image-formats-plugins) package is installed. So the question is: Should lximage-qt advertise support for those formats even when the user will probably get an error if lximage-qt actually tries to open such a file?

tsujan:
> This works because text/x-c...

application/x-mswinurl, application/xml, audio/x-mpegurl, application/rdf+xm, application/x-designer,...,...

> lximage-qt fails to open TIFF and WEBP files unless the qt5-image-format-plugins...

"Recommends" in Debian and its derivatives, "Optional" in Arch and its derivatives, etc. For years, we've repeatedly told users why it's a very bad practice to skip them. No package should compensate for that mistake by fake dependencies.

alexhenrie:
@tsujan, I'm asking you a simple yes-or-no question: Do you think that `image/tiff` and `image/webp` should always be listed in lximage-qt's `.desktop` file?

tsujan:
`MimeType` is good as it is.

alexhenrie:
@tsujan: Thank you for the direct response. So currently, lximage-qt's `MimeType` property has `image/tiff` but not `image/webp`. Why is TIFF okay but not WEBP?

tsujan:
→ https://github.com/lxqt/lximage-qt/issues/592

PR #381, opened 2021-01-03
Implement force zoom fit when loading images.

### LINKED ISSUE #380: An option to enforce fitting on showing next image
IF you show an image from an image dir, zoom it (in or out) then move to next one, zoom settings are preserved instead of reloading initial ones (original vs fit). I don't think this is the expected behavior. You can notice it a lot if you zoom in at max an image and then go to the next one.

##### Expected Behavior

If you like fitting your images, every image you load should be fitted regardless of the zoom (in or out) you did in the previous image.

##### Current Behavior

zoom settings are preserved when moving to a new image

##### Possible Solution

An option in the settings to show new images with zoom to original/zoom to fit settings.
I can work on it if you like the idea.

adjamhub:
Should fix #380

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -572,27 +572,19 @@ void MainWindow::onImageLoaded() {
 
     loadJob_ = nullptr; // the job object will be freed later automatically
 
-    // set image zoom
-    Application* app = static_cast<Application*>(qApp);
-    Settings& settings = app->settings();
-    
-    if (settings.forceZoomFit()) {
+    // set image zoom, like in loadImage()
+    if(static_cast<Application*>(qApp)->settings().forceZoomFit()) {
       ui.actionZoomFit->setChecked(true);
-      ui.view->setAutoZoomFit(true);
-      ui.view->zoomFit();
     }
-    else {
-      // like in onImageLoaded()
-      ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
-      if(ui.actionOriginalSize->isChecked()) {
-        ui.view->zoomOriginal();
-      }
+    ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
+    if(ui.actionOriginalSize->isChecked()) {
+      ui.view->zoomOriginal();
     }
     
     ui.view->setImage(image_);
 
-   // currentIndex_ should be corrected after loading
-   currentIndex_ = indexFromPath(currentFile_);
+    // currentIndex_ should be corrected after loading
+    currentIndex_ = indexFromPath(currentFile_);
 
     updateUI();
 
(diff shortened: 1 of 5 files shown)


stefonarch:
Works fine here!

src/preferencesdialog.ui
```diff
@@ -6,8 +6,8 @@
    <rect>
     <x>0</x>
     <y>0</y>
-    <width>400</width>
-    <height>400</height>
+    <width>427</width>
+    <height>409</height>
    </rect>
```
tsujan, line 11: Please don't change the above sizes when you edit a ui file. I know that Designer did it, based on your Qt style and font sizes but please restore them after using it.

The real sizes will be set by Qt when the app is used and based on user's GUI settings.

src/mainwindow.cpp
```diff
@@ -780,12 +793,23 @@ void MainWindow::loadImage(const Fm::FilePath & filePath, QModelIndex index) {
...
+  }
+  else {
     // like in onImageLoaded()
     ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
     if(ui.actionOriginalSize->isChecked()) {
       ui.view->zoomOriginal();
     }
-
+  }
+
```
tsujan, line 812: Please also replace these lines with:

```c++
    // like in onImageLoaded()
    if(static_cast<Application*>(qApp)->settings().forceZoomFit()) {
      ui.actionZoomFit->setChecked(true);
    }
    ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
    if(ui.actionOriginalSize->isChecked()) {
      ui.view->zoomOriginal();
    }
```

src/mainwindow.cpp
```diff
@@ -572,10 +572,23 @@ void MainWindow::onImageLoaded() {
...
+      ui.view->zoomFit();
     }
+    else {
+      // like in onImageLoaded()
+      ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
+      if(ui.actionOriginalSize->isChecked()) {
+        ui.view->zoomOriginal();
+      }
+    }
+
```
tsujan, line 591: `ImageView::zoomFit` is always called in `imageview.cpp` when an image is set. Since the image is set below these lines, we don't need to call it and the above code can be simplified as:

```c++
    if(static_cast<Application*>(qApp)->settings().forceZoomFit()) {
      ui.actionZoomFit->setChecked(true);
    }
    ui.view->setAutoZoomFit(ui.actionZoomFit->isChecked());
    if(ui.actionOriginalSize->isChecked()) {
      ui.view->zoomOriginal();
    }
```

src/preferencesdialog.ui
adjamhub: I'll take care about here and the next times :)

adjamhub:
I hope I applied well your suggestions :)

tsujan (approved):
GTM.

Thanks!

PR #397, opened 2021-03-03
Added options for thumbnails

tsujan:
Two options are added: showing thumbnails dock by default and setting the max. file size for creating thumbnails.

No option is added for disabling thumbnails because this is an image viewer.

Also, silenced two compilation warnings about `QNetworkReply::error` and `QMatrix` and changed a bad translatable string that was added in https://github.com/lxqt/lximage-qt/commit/502530154d61d9e46213908292ecc8a7c0e0520c


stefonarch:
I was a little confused by the fact that checking the box for showing toolbar always didn't have an immediate effect (when  View > show thumbnails is unchecked).

tsujan:
> I was a little confused by the fact that checking the box for showing toolbar always didn't have an immediate effect

Yes, it was only for the current session. With the current patch, it'll still be for the current session but the user could choose to show the thumbnail doc on startup by using Preferences. I think that's more flexible.

PR #398, opened 2021-03-11
Fix all compilation warnings and remove deprecated methods

adjamhub:


tsujan:
Was in conflict with https://github.com/lxqt/lximage-qt/pull/397. Please rebase or make another PR.

adjamhub:
Rebased!
The conflicts were due to similar fixes with deprecated methods :)

src/imageview.cpp
```diff
@@ -98,7 +98,9 @@ void ImageView::onFileDropped(const QString file) {
 }
 
 void ImageView::wheelEvent(QWheelEvent* event) {
-  int delta = event->delta();
+  QPoint angleDelta = event->angleDelta();
+  Qt::Orientation orient = (qAbs(angleDelta.x()) > qAbs(angleDelta.y()) ? Qt::Horizontal : Qt::Vertical);
+  int delta = (orient == Qt::Horizontal ? angleDelta.x() : angleDelta.y());
```
tsujan, line 103: I think Qt should restore `QWheelEvent::delta()` or add another method similar to it. Each time we need delta, we should do this calculation to get rid of warnings :(

tsujan (approved):
GTM.

Thanks!

src/imageview.cpp
adjamhub: Definitely!
I can try asking in some Qt chats about...

PR #438, opened 2021-09-01
Add option to hide/show main toolbar

xiota:
It's currently possible to hide the main toolbar using its context menu.  However, to re-show it requires using the context menu from the annotation toolbar.  This makes re-showing the main toolbar a multi-step process (show annotation bar, context menu, show main toolbar, re-hide annotation bar).  This pull request adds the option to hide/show the main toolbar directly.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -117,6 +117,20 @@ MainWindow::MainWindow():
   setShowThumbnails(settings.showThumbnails());
   ui.actionShowThumbnails->setChecked(settings.showThumbnails());
 
+  ui.toolBar->setVisible(settings.isToolbarShown());
+  ui.actionToolbar->setChecked(settings.isToolbarShown());
+  connect(ui.actionToolbar, &QAction::triggered, this, [this](int checked) {
+    if(!isFullScreen()) { // toolbar is hidden in fullscreen
+      ui.toolBar->setVisible(checked);
+    }
+  });
+  // toolbar visibility can change in its context menu
+  connect(ui.toolBar, &QToolBar::visibilityChanged, this, [this](int visible) {
+    if(!isFullScreen()) { // toolbar is hidden in fullscreen
+      ui.actionToolbar->setChecked(visible);
+    }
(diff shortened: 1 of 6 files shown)


tsujan (approved):
GTM. Thanks!

@stefonarch, would you please test it too? The behavior is like the annotation toolbar: showing/hiding of the main toolbar shouldn't be remembered but there's an option for it in Preferences dialog.

stefonarch:
> (show annotation bar, context menu, show main toolbar, re-hide annotation bar)

Maybe I'm missing something, but to me steps are : right click besides menu texts to show context menu > select show toolbar and/or annotation bar:

schermata-09-02-07-37-01

So I cannot notice any change now...

xiota:
I'm using a global menubar, so I don't have that context menu on the menubar.  Also, some people might like the toolbar hidden by default on startup (option in preferences).

xiota:
I made some more changes to preferences.  Changed the "Thumbnails" tab to "Dock".  Added options to show Exif data on startup, change position of the Thumbnails dock, and change the size of thumbnails.   Should I add it to this PR, or open up a new one?  Here's a screenshot:

screenshot-01-dock

tsujan:
> Should I add it to this PR

Please don't do that! Each PR should be about a single change.

tsujan:
@stefonarch 
Yes, the toolbar can be shown/hidden by right clicking on the menu-bar too but I see this PR as a way of treating the main toolbar like the annotation toolbar and having an option for hiding it with startup.

xiota:
@tsujan So... I should submit a separate pull request for each option?

tsujan:
>  I should submit a separate pull request for each option?

Yes, please. Each PR should be about a single subject. The current PR is good as it is. It's about the main toolbar.

Please also note that the second tab of Preferences is only about thumbnails. I don't think we should add exif data to it but we could discuss it when you make your next PR.

tsujan:
I merge this because its code is clean and it treats both toolbars in the same way.

PR #439, opened 2021-09-02
Add option to use system Trash

xiota:
Add an option to use the system trash instead of immediately deleting files.

screenshot-trash

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -491,9 +491,15 @@ void MainWindow::on_actionSaveAs_triggered() {
 }
 
 void MainWindow::on_actionDelete_triggered() {
-  // delete the current file
-  if(currentFile_)
-    Fm::FileOperation::deleteFiles({currentFile_});
+  // delete or trash the current file
+  if(currentFile_) {
+    if(static_cast<Application*>(qApp)->settings().useTrash()) {
+      Fm::FileOperation::trashFiles({currentFile_}, false);
+    }
+    else {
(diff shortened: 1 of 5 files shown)


src/preferencesdialog.ui
```diff
@@ -117,6 +117,13 @@
          </property>
         </widget>
        </item>
+       <item row="9" column="0" colspan="2">
+        <widget class="QCheckBox" name="useTrashBox">
+         <property name="text">
+          <string>Use system Trash (and don't prompt)</string>
```
tsujan, line 123: Very good! Please just change "don't" to "do not".

xiota: Done.

src/mainwindow.cpp
```diff
@@ -491,9 +491,14 @@ void MainWindow::on_actionSaveAs_triggered() {
...
 
 void MainWindow::on_actionDelete_triggered() {
-  // delete the current file
-  if(currentFile_)
-    Fm::FileOperation::deleteFiles({currentFile_});
+  // delete or trash the current file
+  if(currentFile_) {
+    if(static_cast<Application*>(qApp)->settings().useTrash()) {
+      Fm::FileOperation::trashFiles({currentFile_}, false);
+    } else {
```
tsujan, line 498: Please also respect the original code style here:

```c++
    }
    else {
```

xiota: done.

xiota:
sorry, got branches mixed up... hold on.

tsujan (approved):
Thanks a lot! GTM.

PR #440, opened 2021-09-02
Add option to change Thumbnail dimensions

xiota:
Add option in preferences to set preferred thumbnail size.

screenshot-thumbsize-combo

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1205,7 +1205,8 @@ void MainWindow::setShowThumbnails(bool show) {
       thumbnailsDock_->setFeatures(QDockWidget::NoDockWidgetFeatures); // FIXME: should use DockWidgetClosable
       thumbnailsDock_->setWindowTitle(tr("Thumbnails"));
       thumbnailsView_ = new Fm::FolderView(Fm::FolderView::IconMode);
-      thumbnailsView_->setIconSize(Fm::FolderView::IconMode, QSize(64, 64));
+      Settings& settings = static_cast<Application*>(qApp)->settings();
+      thumbnailsView_->setIconSize(Fm::FolderView::IconMode, QSize(settings.thumbnailSize(), settings.thumbnailSize()));
       thumbnailsView_->setAutoSelectionDelay(0);
       thumbnailsDock_->setWidget(thumbnailsView_);
       addDockWidget(Qt::BottomDockWidgetArea, thumbnailsDock_);
--- a/src/preferencesdialog.cpp
+++ b/src/preferencesdialog.cpp
@@ -91,6 +91,7 @@ PreferencesDialog::PreferencesDialog(QWidget* parent):
   ui.thumbnailBox->setChecked(settings.showThumbnails());
   // the max. thumbnail size spinbox is in MiB
   ui.thumbnailSpin->setValue(qBound(0, settings.maxThumbnailFileSize() / 1024, 1024));
+  initThumbnailSizes(settings);
 
   // shortcuts
   initShortcuts();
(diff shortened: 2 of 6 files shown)


tsujan:
Having separate widths and heights would be too much and could lead to complications. I suggest we change the text to "Size of thumbnails" (as it is in pcmanfm-qt's Preferences) and use a a combo-box instead of the spin-box, with sizes 64, 96, 128, 160, 192, 224, 256. See pcmanfm-qt's code and, especially, https://github.com/lxqt/pcmanfm-qt/blob/fa3fb863478e97283532abd9f85925c9b210de4d/pcmanfm/settings.cpp#L491

xiota:
What do you mean by "too much"?  What types of complications are you thinking of?  256 might not be enough for high dpi monitors.  It's fine for icons in a file manager, but this is an image viewer.

tsujan:
By "too much" I meant an overdoing. We don't do that elsewhere. Also, please note that not all images can be thumbnailed and so, icons may be shown too.

> 256 might not be enough for high dpi monitors

We don't support more than 256 in the core of `libfm-qt`. That might change later but, until then, (1) we prefer a unified approach in LXQt apps, and (2) users with HDPI monitors can use 128; libfm-qt considers the aspect ratio.

xiota:
okay.  I'll work on it.

tsujan:
There's no rush. Happy to see that you add useful features to lximage-qt in a clean way!

xiota:
@tsujan It's changed to a combobox.  I updated the screenshot in the main description.

tsujan:
Looks good. I'll review it tomorrow.

tsujan:
Adding size strings to `preferencesdialog.ui` → `thumbnailSizeComboBox` makes the code less flexible and more error-prone. A cleaner approach is like what is done in `pcmanfm-qt`:

1. `thumbnailSize_` should be an integer, with 64 as the default.

2. A static function is needed in `settings.h`:

```c++
static const QList<int> & thumbnailSizes();
```

With this definition in `settings.cpp`:

```c++
const QList<int>& Settings::thumbnailSizes() {
  static const QList<int> _thumbnailSizes = {256, 224, 192, 160, 128, 96, 64};
  return _thumbnailSizes;
}
```

3. The combo-box can be filled in by using something like this (the smaller sizes come lower):

```c++
int i = 0;
for(const auto & size : Settings::thumbnailSizes()) {
  ui.thumbnailSizeComboBox->addItem(QStringLiteral("%1 x %1").arg(size), size);
  if(settings.thumbnailSize() == size) {
    ui.thumbnailSizeComboBox->setCurrentIndex(i);
  }
  ++i;
}
```

Then, the size can be known by using `QComboBox::itemData()`.

xiota:
@tsujan  Thanks for explaining.  I've made changes according to what you've suggested.  Would you want similar changes for the thumbnail dock position, or would keeping those as strings be okay?

src/mainwindow.cpp
```diff
@@ -1199,13 +1199,15 @@ void MainWindow::on_actionShowExifData_triggered(bool checked) {
 }
 
 void MainWindow::setShowThumbnails(bool show) {
+  Settings& settings = static_cast<Application*>(qApp)->settings();
```
tsujan, line 1202: For now, `settings` is used only in one place inside this function. Logically, it's better to put it before the line, where it's used (although there's no practical change).

xiota: It will be used more in the next PR (to choose the dock position).  Would you still want it moved back and forth between the PRs?

xiota: I went ahead and moved it.

tsujan (approved):
GTM

tsujan:
Thanks for your work! I'm merging this....

I know that you have more plans for lximage-qt and I've found them interesting but I can't consider a future PR unless the current one is reviewed, approved and merged.

PR #441, opened 2021-09-02
Add option to change Thumbnail dock position

xiota:
This adds an option to move the Thumbnail dock to different sides of the main window.

screenshot-thumb-position-2

### DIFF
--- a/src/preferencesdialog.cpp
+++ b/src/preferencesdialog.cpp
@@ -92,6 +92,7 @@ PreferencesDialog::PreferencesDialog(QWidget* parent):
   // the max. thumbnail size spinbox is in MiB
   ui.thumbnailSpin->setValue(qBound(0, settings.maxThumbnailFileSize() / 1024, 1024));
   initThumbnailSizes(settings);
+  initThumbnailsPositions(settings);
 
   // shortcuts
   initShortcuts();
@@ -151,6 +152,7 @@ void PreferencesDialog::accept() {
   settings.setUseTrash(ui.useTrashBox->isChecked());
 
   settings.setShowThumbnails(ui.thumbnailBox->isChecked());
+  settings.setThumbnailsPosition(ui.thumbnailsPositionComboBox->currentText());
   // the max. thumbnail size spinbox is in MiB
   settings.setMaxThumbnailFileSize(ui.thumbnailSpin->value() * 1024);
   settings.setThumbnailSize(ui.thumbnailSizeComboBox->itemData(ui.thumbnailSizeComboBox->currentIndex()).toInt());
@@ -191,6 +193,13 @@ void PreferencesDialog::initThumbnailSizes(Settings& settings) {
     }
     ++i;
   }
+}
+
+void PreferencesDialog::initThumbnailsPositions(Settings& settings) {
+  for (auto position : settings.thumbnailsPositions()) {
+    ui.thumbnailsPositionComboBox->addItem(position);
+  }
+  ui.thumbnailsPositionComboBox->setCurrentText(settings.thumbnailsPosition());
 }
 
 void PreferencesDialog::initIconThemes(Settings& settings) {
--- a/src/settings.cpp
+++ b/src/settings.cpp
@@ -31,6 +31,7 @@ Settings::Settings():
   fullScreenBgColor_(0, 0, 0),
   showThumbnails_(false),
   thumbnailSize_(64),
+  thumbnailsPosition_(QStringLiteral("bottom")),
   showSidePane_(false),
   slideShowInterval_(5),
   fallbackIconTheme_(QStringLiteral("oxygen")),
@@ -86,8 +87,9 @@ bool Settings::load() {
 
   settings.beginGroup(QStringLiteral("Thumbnail"));
   showThumbnails_ = settings.value(QStringLiteral("ShowThumbnails"), false).toBool();
+  setThumbnailsPosition(settings.value(QStringLiteral("ThumbnailsPosition")).toString());
   setMaxThumbnailFileSize(qMax(settings.value(QStringLiteral("MaxThumbnailFileSize"), 4096).toInt(), 1024));
-  thumbnailSize_ = settings.value(QStringLiteral("ThumbnailSize"), 64).toInt();
+  setThumbnailSize(settings.value(QStringLiteral("ThumbnailSize"), 64).toInt());
   settings.endGroup();
 
   return true;
@@ -132,6 +134,7 @@ bool Settings::save() {
 
   settings.beginGroup(QStringLiteral("Thumbnail"));
   settings.setValue(QStringLiteral("ShowThumbnails"), showThumbnails_);
+  settings.setValue(QStringLiteral("ThumbnailsPosition"), thumbnailsPosition_);
   settings.setValue(QStringLiteral("MaxThumbnailFileSize"), maxThumbnailFileSize());
   settings.setValue(QStringLiteral("ThumbnailSize"), thumbnailSize_);
   settings.endGroup();
--- a/src/settings.h
+++ b/src/settings.h
@@ -71,6 +71,37 @@ public:
     showThumbnails_ = show;
   }
 
+  QString thumbnailsPosition() {
+    return thumbnailsPosition_;
+  }
+  void setThumbnailsPosition(const QString& position) {
+    if(!thumbnailsPositions().contains(position)) {
+      thumbnailsPosition_ = QStringLiteral("bottom");
+    }
+    else {
+      thumbnailsPosition_ = position;
+    }
+  }
+  const QStringList& thumbnailsPositions() {
+    static const QStringList _thumbnailsPositions = {QStringLiteral("left"),
+                                                     QStringLiteral("right"),
+                                                     QStringLiteral("top"),
+                                                     QStringLiteral("bottom")};
+    return _thumbnailsPositions;
+  }
+  Qt::DockWidgetArea getThumbnailsPosition() {
+    if(thumbnailsPosition_ == QStringLiteral("left")) {
+      return Qt::LeftDockWidgetArea;
+    }
+    if(thumbnailsPosition_ == QStringLiteral("right")) {
+      return Qt::RightDockWidgetArea;
+    }
+    if(thumbnailsPosition_ == QStringLiteral("top")) {
+      return Qt::TopDockWidgetArea;
+    }
+    return Qt::BottomDockWidgetArea;
+  }
+
   int maxThumbnailFileSize() const {
     return Fm::ThumbnailJob::maxThumbnailFileSize();
   }
@@ -87,7 +118,12 @@ public:
     return thumbnailSize_;
   }
   void setThumbnailSize(int size) {
-    thumbnailSize_ = size;
+    if(!thumbnailSizes().contains(thumbnailSize_)) {
+      thumbnailSize_ = 64;
+    }
+    else {
+      thumbnailSize_ = size;
+    }
   }
 
   bool showSidePane() const {
@@ -236,6 +272,7 @@ private:
   QColor fullScreenBgColor_;
   bool showThumbnails_;
   int thumbnailSize_;
+  QString thumbnailsPosition_;
   bool showSidePane_;
   int slideShowInterval_;
   QString fallbackIconTheme_;
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -659,25 +659,6 @@ bool MainWindow::eventFilter(QObject* watched, QEvent* event) {
       default:;
     }
   }
-  else if(thumbnailsView_ && watched == thumbnailsView_->childView()) {
-    // scroll the thumbnail view with mouse wheel
-    switch(event->type()) {
-      case QEvent::Wheel: { // mouse wheel event
-        QWheelEvent* wheelEvent = static_cast<QWheelEvent*>(event);
-        if(wheelEvent->modifiers() == 0) {
-          QPoint angleDelta = wheelEvent->angleDelta();
-          Qt::Orientation orient = (qAbs(angleDelta.x()) > qAbs(angleDelta.y()) ? Qt::Horizontal : Qt::Vertical);
-          int delta = (orient == Qt::Horizontal ? angleDelta.x() : angleDelta.y());
-          QScrollBar* hscroll = thumbnailsView_->childView()->horizontalScrollBar();
-          if(hscroll)
-            hscroll->setValue(hscroll->value() - delta);
-          return true;
-        }
-        break;
-      }
-      default:;
-    }
-  }
   return QObject::eventFilter(watched, event);
 }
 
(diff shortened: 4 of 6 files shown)


tsujan:
This will be the next PR to be reviewed. Please rebase!

To avoid conflicts, you could wait for your last PR to be reviewed/merged before making another.

xiota:
@tsujan Rebased.  I already have some pull requests opened, but won't make anymore until they're done being reviewed.

tsujan:
> I already have some pull requests opened, but won't make anymore until they're done being reviewed.

It was just a suggestion. Personally, I prefer to avoid rebasing with my PRs but you may be more comfortable with it. In the end, the difference is only a rebase request.

xiota:
@tsujan I opened them in case you were interested in seeing where this is heading.  The next PR, adding showing/hiding the Exif dock on startup is where it's relevant because the General tab is getting cluttered.  If you know where it's heading, you might have a better idea of how it should be organized.

tsujan:
> I opened them in case you were interested in seeing where this is heading.

I am interested — I usually don't review PRs so quickly — and it's good to see that you make each PR only for one change. Having a limited time, I should focus on each PR separately and overlook the rest *temporarily* for a clean review.

xiota:
Thank you for taking the time to review.  I'm isolating the PRs to individual changes because you mentioned that preference before.

tsujan:
Testing the PR, I encountered a strange thing that wasn't visible before: When the thumbnail dock is vertical, the wheel scroll lines isn't respected. By default, it's three, so *exactly* three items should be scrolled.

Therefore, I removed the filter event for `thumbnailsView_->childView()` altogether and everything was fine. I think the filter was used as a workaround for a problem in old Qt versions. Please remove the filter *offline* and test the result! Tell me if you see the same thing that I see. Then we could decide whether we should remove it from the PR.

There's a difference between wheel scrolling in vertical and horizontal docks that isn't related to the current PR: the former is smooth, while the latter is not. The reason is that the existence of a horizontal scrollbar isn't predicted in `libfm-qt`. I'll try to fix it later but, as far as this PR is concerned, I think we can just remove the filter.

xiota:
I noticed some strange behavior with scrolling, but thought I've been having problems with my touchpad with other apps (eg, Firefox).  I'll look for a mouse with a real scroll wheel while it's building to test what you describe.

xiota:
I can't find a mouse with a scrollwheel.  On my computer, the scrolling becomes unusably glitchy without the `thumbnailsView_->childView()` filter event.  The other issue with scrolling that I mentioned earlier is when scrolling on the main image, the images flip by too quickly for scrolling to be useful (makes me want to add an option to disable scrolling) .

When you say it doesn't respect the scroll lines, is it scrolling too much or too little?  And when you say "three items" is that number of thumbnail images or an amount that would normally correspond with three lines (as if the scrollable area had lines of text instead of thumbnails)?

tsujan:
> The other issue with scrolling that I mentioned earlier is when scrolling on the main image

Let's forget about image scrolling and focus on the thumbnail dock, which is the main topic of the PR.

> When you say it doesn't respect the scroll lines, is it scrolling too much or too little?

With the vertical dock and 128-pix thumbnails, it's a little less than three images. "Line" actually means "items" (= images and their texts). Without the filter, it's exactly three items.

Actually, at least for the vertical dock, no filtering is needed because it's done by `libfm-qt`. Let me do some experiments with `libfm-qt` first; then I'll come back and tell you. It may take a while.

tsujan:
OK, my experiments were much shorter than I thought. My guess was correct: the filtering was done because of an old bug in Qt. It's fixed now.

Please remove the block `else if(thumbnailsView_ && watched == thumbnailsView_->childView()) {` from `MainWindow::eventFilter` (alongside `Settings& settings = static_cast<Application*>(qApp)->settings()`) and also `listView->installEventFilter(this);` from `MainWindow::setShowThumbnails`.

I'll make a patch for `libfm-qt` soon. It'll make the wheel scrolling smooth for the horizontal dock too and, with the above-mentioned change, not only the number of scrolled items will be correct everywhere but also the user could enjoy `Shift + mouse-wheel` in the horizontal dock (= one-item scrolling).

xiota:
Okay...  BTW, what's the difference between `IconMode` and `ThumbnailMode`.  Should the thumbnail dock use `ThumbnailMode` ?

tsujan:
> BTW, what's the difference between IconMode and ThumbnailMode

`ThumbnailMode` exists only to be distinguished from `IconMode` by apps that need both (like pcmanfm-qt). In lximage-qt, there's no difference and `IconMode` is good.

xiota:
I've pushed the changes.

tsujan:
The attached screencast shows smooth scrolling with the horizontal dock (after a change in `libfm-qt`).

https://user-images.githubusercontent.com/981076/132043920-dc672244-50ef-4534-8bb8-352b9384bf06.mp4

xiota:
Vertical scrolling works fine in pcmanfm-qt, but is glitchy in lximage-qt without the filter.  But it can be looked at later.

tsujan:
> Vertical scrolling works fine in pcmanfm-qt, but is glitchy in lximage-qt without the filter. 

Here, it's fine with both mouse wheel and touchpad.

As for scrolling when the cursor is inside the image, I tested with touchpad and saw that it wasn't usable. We should find the cause but not in this PR.

xiota:
> We should find the cause but not in this PR.

Yes, another PR.

tsujan:
Mouse wheel scrolling with the vertical dock (after a change in `libfm-qt` but I don't think that change is relevant here):


https://user-images.githubusercontent.com/981076/132046491-adf59c23-3998-4d6b-adc7-598133d5a1bd.mp4

xiota:
The problem is... after the first jump, scrolling just freezes with my touchpad.  But I think it can wait till I get a chance to try out the libfm-qt patch.  If that doesn't fix it, I can always just add a "smooth scroll" option.  Not everyone necessarily wants it to jump three images at a time.

xiota:
I don't know how to record a screen capture.  What program are you using?

tsujan:
> The problem is... after the first jump, scrolling just freezes with my touchpad.

Very strange! I can't reproduce it here. Moreover, since we removed the filter, it should be exactly as it is in pcmanfm-qt.

> I don't know how to record a screen capture. What program are you using?

SimpleScreenRecorder. But there's no need to record it; I believe you. I just don't understand how it's possible in your case.

BTW, I opened https://github.com/lxqt/lximage-qt/issues/444

xiota:
I don't know why it isn't like in pcmanfm-qt.  I'm willing to wait to see if the change to libfm-qt makes a difference.  If not, I can open an issue with a screen capture so you can see what it looks like.

tsujan:
> I'm willing to wait to see if the change to libfm-qt makes a difference.

Oh, there's a possibility: in the `libfm-qt` patch, I also considered horizontal angle data, while it was ignored before (it was considered in lximage-qt though).

xiota:
The first half is with the trackball.  The second half with the touchpad.

https://user-images.githubusercontent.com/8353098/132050178-42bf4064-f2c1-453b-ab49-1f1c569962fc.mp4

xiota:
libfm-qt on my computer is 0.17.1 built from the debian source package.  I'll try building it from git.

xiota:
Distros tend to lag behind in versions... so it might be worth thinking about adding a smooth scroll option in a future PR.

tsujan:
> libfm-qt on my computer is 0.17.1 built from the debian source package. I'll try building it from git.

I haven't made a PR yet. So, it'll make no difference.

Some touchpads distinguish between horizontal and vertical angle data, some don't (I have both). In the first case, the patch that I'll add as a PR will make a difference.

xiota:
I have horizontal scrolling turned off on my touchpad.  With it on, at least without the patch you're working on, it makes no difference.  I just tried touch screen too, and it picks up to drag the icon.  Would be nice if it could scroll with touch.  (Feature for another PR...)

tsujan:
Apart from all these, your screencast shows a weird situation. With touchpad, the scrolling should be smooth. Does your touchpad have a setting for disabling it?

xiota:
Disable when typing?  It was turned off (no disabling) when I checked earlier to try enabling horizontal.  Turning it on and off makes no difference.

I just build libfm-qt from git and the thumbnails don't change size anymore...  is that something to try to figure out now or later?

tsujan:
> Disable when typing?

No. Maybe the frame rate of your video wasn't good but it shows non-smooth (jumpy) scrolling with touchpad, which shouldn't be possible.

> I just build libfm-qt from git and the thumbnails don't change size anymore

When you upgrade libfm-qt, you should upgrade and recompile all apps that are based on it and then restart them; otherwise, they might not work at all.

xiota:
That's how scrolling is behaving.  It freezes up in the middle of scrolling.

I did rebuild and restart lximage.  I just rebooted and am rebuilding lximage again.  Same problem as earlier with thumbnail sizes.  I'm going to remove the git libfm-qt.

tsujan:
> I'm going to remove the git libfm-qt.

It has no problem. I use nothing but it. Somewhere in your tests (compilation, building,...), there should be a mistake.

xiota:
I'll figure it out later.  It's a different issue...

If you're trying to avoid a regression, the smooth scrolling should probably stay in until you push out the libfm-qt changes.  Otherwise, I don't mind that scrolling doesn't work for me.  It's something I can deal with later.  There's also the main window scrolling issue that has to be fixed later.

tsujan:
> If you're trying to avoid a regression, the smooth scrolling should probably stay in

I'm feeling that we may have a mutual misunderstanding here, especially because of the touchpad problem you see.

Anyway, the patch is here: https://github.com/lxqt/libfm-qt/pull/721. You need the git libfm-qt for trying it.

As for compiling git libfm-qt, Debian isn't the best distro for that but it's quite possible with it (I used Debian before). First of all, you should follow compilation order and, to rule out any issue, you need to upgrade the whole LXQt to git (or parts of it that you use) in the above-mentioned order. Second, you need take care of package splitting in Debian (libfm-qt, libfm-qt-dev, etc.).

xiota:
What I'm wondering is... in terms of this PR... is there anything else to do?  I've already removed the smooth scrolling (event capture) bits.

The only parts of lxqt that I use are pcmanfm-qt.  I haven't used lximage-qt before because it's missing features I think are essential, which is basically what I'm adding in one-by-one.

What do you mean by "clean deb packages"?

tsujan:
> What I'm wondering is... in terms of this PR... is there anything else to do?

I just tested it; haven't reviewed it yet ;)

> What do you mean by "clean deb packages"?

Like deb packages Debian provides but for git, so that they can overwrite the latter cleanly and the package manager knows about the upgrade.

xiota:
Rebuilding and installing libfm-qt as a deb package seems to work fine.  I'll try out the patch now.

xiota:
Build libfm-qt with the new patch.  Same issue as before.  Something I didn't notice before though.  If my mouse cursor is over the scrollbar, scrolling works perfectly.  It's only when it's over the thumbnails themselves that it freezes up.

tsujan:
This is the vertical dock with touchpad (see how smoothly I can scroll it):


https://user-images.githubusercontent.com/981076/132057962-0e7afc71-c9fc-4f94-b290-5b717d6be1dd.mp4

I don't think the problem you see is related to this PR. You could open an issue after it's merged.

xiota:
> This is the vertical dock with touchpad (see how smoothly I can scroll it):

Haha.  I get that it works fine on your computer...

> I don't think the problem you see is related to this PR. You could open an issue after it's merged.

I don't mind opening an issue later if needed.

tsujan:
> Haha. I get that it works fine on your computer...

The problem is stranger than that because you say you don't see a problem in pcmanfm-qt, but now, lximage-qt is exactly like pcmanfm-qt when its thumbnail dock is vertical.

> I actually put this PR before the next one because I thought (this one) would require less discussion...

It can't be predicted with PRs. I've had PRs that were straightforward to me but were followed by pertinent, long discussions.

xiota:
So libfm-qt has a rename dialog.  I'd like to use it in lximage-qt, but I don't know how...  Is there any example code that uses?

xiota:
I tried with another computer running KDE.  Same vertical scrolling issue.  The video I sent earlier was captured on Xfce.

src/settings.h
```diff
@@ -71,6 +71,24 @@ class Settings {
...
+    static const QMap<QString, Qt::DockWidgetArea> _thumbnailsPositions = {
+        {QStringLiteral("left"), Qt::LeftDockWidgetArea},
+        {QStringLiteral("right"), Qt::RightDockWidgetArea},
+        {QStringLiteral("top"), Qt::TopDockWidgetArea},
+        {QStringLiteral("bottom"), Qt::BottomDockWidgetArea} };
+    return _thumbnailsPositions;
+  }
+  Qt::DockWidgetArea getThumbnailsPosition() {
+    return thumbnailsPositions()[thumbnailsPosition_];
+  }
```
tsujan, line 90: `QMap` isn't needed here:

```c++
  QString thumbnailsPosition() {
    return thumbnailsPosition_;
  }
  void setThumbnailsPosition(const QString& position) {
    if(!thumbnailsPositions().contains(position)) {
      thumbnailsPosition_ = QStringLiteral("bottom");
    }
    else {
      thumbnailsPosition_ = position;
    }
  }
  const QStringList& thumbnailsPositions() {
    static const QStringList _thumbnailsPositions = {QStringLiteral("left"),
                                                     QStringLiteral("right"),
                                                     QStringLiteral("top"),
                                                     QStringLiteral("bottom")};
    return _thumbnailsPositions;
  }
  Qt::DockWidgetArea getThumbnailsPosition() {
    if(thumbnailsPosition_ == QStringLiteral("left")) {
      return Qt::LeftDockWidgetArea;
    }
    if(thumbnailsPosition_ == QStringLiteral("right")) {
      return Qt::RightDockWidgetArea;
    }
    if(thumbnailsPosition_ == QStringLiteral("top")) {
      return Qt::TopDockWidgetArea;
    }
    return Qt::BottomDockWidgetArea;
  }
```

It also guards against meaningless strings in the config file (which can be edited manually).

It's preferable to put function definitions in `settings.cpp`.

src/settings.cpp
```diff
@@ -86,6 +87,7 @@ bool Settings::load() {
 
   settings.beginGroup(QStringLiteral("Thumbnail"));
   showThumbnails_ = settings.value(QStringLiteral("ShowThumbnails"), false).toBool();
+  thumbnailsPosition_ = settings.value(QStringLiteral("ThumbnailsPosition"), QStringLiteral("bottom")).toString();
```
tsujan, line 90: Please change this to (see my other comment for the reason):

```c++
  thumbnailsPosition_ = settings.value(QStringLiteral("ThumbnailsPosition")).toString();
  if(!thumbnailsPositions().contains(thumbnailsPosition_)) {
    thumbnailsPosition_ = QStringLiteral("bottom");
  }
```

src/preferencesdialog.cpp
```diff
@@ -193,6 +195,13 @@ void PreferencesDialog::initThumbnailSizes(Settings& settings) {
   }
 }
 
+void PreferencesDialog::initThumbnailsPositions(Settings& settings) {
+  for (auto position : settings.thumbnailsPositions().toStdMap()) {
+    ui.thumbnailsPositionComboBox->addItem(position.first);
+  }
+  ui.thumbnailsPositionComboBox->setCurrentText(settings.thumbnailsPosition());
+}
+
```
tsujan, line 204: If we consider the previous comments, this needs to be changed to:

```c++
void PreferencesDialog::initThumbnailsPositions(Settings& settings) {
  for(auto position : settings.thumbnailsPositions()) {
    ui.thumbnailsPositionComboBox->addItem(position);
  }
  ui.thumbnailsPositionComboBox->setCurrentText(settings.thumbnailsPosition());
}
```

src/settings.h
xiota: I'll make these changes.

I put these function definitions here because every other function of this type in this section is defined here.

tsujan:
To be on the safe side, it's better to do a similar thing for your previous PR too:

```c++
  thumbnailSize_ = settings.value(QStringLiteral("ThumbnailSize")).toInt();
  if(!thumbnailSizes().contains(thumbnailSize_)) {
    thumbnailSize_ = 64;
  }
```

You could do it here, as an exception.

xiota:
You don't want the sizes change in a separate PR?

tsujan:
> You don't want the sizes change in a separate PR?

You mean my last suggestion? No, we can do it here; we'll have lots of work together later.

src/settings.h
tsujan: > I put these function definitions here because...

I have no objection.

xiota:
What if we use the `set___()` functions instead of rewriting the guarding code everywhere?

tsujan:
> What if we use the set___() functions instead of rewriting the guarding code everywhere?

Do it as you see fit — it's *your* PR. I review its code and don't want to force my personal habits.

xiota:
I just pushed the changes.

tsujan (approved):
Very good.

I requested a test about touchpad from @stefonarch. We might discuss it later.

xiota:
I'll rebase the next one.

tsujan:
> I'll rebase the next one.

Good. I'll see if I could review it tomorrow; if not, it won't take long.

xiota:
I separated out the potential GUI reorganization for a different PR if you'd prefer.  Basically, I'm running out of room for options and the General tab is just a hodgepodge of stuff.

tsujan:
Fine with me.

xiota:
Thanks for your help with these.  Will look forward to going over the next one.

tsujan:
Your welcome! And thanks for adding practical features!

xiota:
It's been a pleasure.

PR #442, opened 2021-09-02
Add option to show/hide Exif dock at startup

xiota:
This adds an option to preferences to show/hide the Exif dock at startup.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -114,6 +114,9 @@ MainWindow::MainWindow():
   ui.view->showOutline(settings.isOutlineShown());
   ui.actionShowOutline->setChecked(settings.isOutlineShown());
 
+  setShowExifData(settings.showExifData());
+  ui.actionShowExifData->setChecked(settings.showExifData());
+
   setShowThumbnails(settings.showThumbnails());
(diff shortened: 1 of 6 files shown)


tsujan:
A quick review before I go to sleep:

1. `setShowExifData()` doesn't show the exif dock *explicitly* (and it shouldn't, because of the way it's used). That causes a problem with the changes you've made: if exif dock's action is triggered in the full-screen state, the dock won't be shown. It can be fixed by adding the following lines to the end of `MainWindow::on_actionShowExifData_triggered`:

```c++
  if(checked && exifDataDock_) {
    exifDataDock_->show(); // needed in the full-screen state
  }
```
2. `exifDataDockView_`, `exifDataDockViewContent_` and `exifDataContentTable_` are redundant as private variables and can be removed from `mainwindow.h`. They're declared inside `setShowExifData()`. (This isn't caused by your PR — I just hadn't noticed it before.)

3. The name of the Preferences' second tab isn't good now. Please make another PR after this one is merged, in order to put the options for the visibility of toolbars under the second tab. IMO, "Window" will be a good name for it but, if you know of a better name, use it.

tsujan (approved):
GTM

Issue #444, opened 2021-09-03
Wheel scrolling on the image isn't usable with touchpad


tsujan:
Sorry for the short report! Please just test it and you'll see why it isn't usable: the image changes unpredictably.

Wheel scrolling is OK with mouse though.

PR #446, opened 2021-09-04
Wait for image to load when using scroll wheel

xiota:
Possible fix for #444


tsujan:
Please see https://github.com/lxqt/lximage-qt/pull/448

tsujan:
Superceded by https://github.com/lxqt/lximage-qt/pull/448

PR #447, opened 2021-09-04
 Reorganize preferences dialog 

xiota:
Proposed reorganization of preferences.

screenshot-08-preferences

### DIFF
--- a/src/preferencesdialog.ui
+++ b/src/preferencesdialog.ui
@@ -19,11 +19,11 @@
      <property name="currentIndex">
       <number>0</number>
      </property>
-     <widget class="QWidget" name="tab">
+     <widget class="QWidget" name="tab_1">
       <attribute name="title">
        <string>General</string>
       </attribute>
-      <layout class="QFormLayout" name="formLayout">
+      <layout class="QFormLayout" name="formLayout_1">
        <property name="fieldGrowthPolicy">
         <enum>QFormLayout::AllNonFixedFieldsGrow</enum>
        </property>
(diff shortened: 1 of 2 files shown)


tsujan:
Please:

 1. Don't touch `settings.cpp`! Changing groups is backward incompatible and we avoid that in LXQt. Moreover, there's no need to a correspondence between Preferences dialog and config file. The former is for the user, while the latter is for the app.
 2. Don't drop "by default". It means "when the app starts". The changes made to the main window are temporary and that's a feature we don't want to remove.

xiota:
I put `settings.cpp` back to how it was before except that I moved `ShowExifData` to the `Windows` section.  It was in the `Thumbnails` section before, and the people this change affects should be limited since it was added only recently.

I put "by default" back.  It just seemed redundant because the preferences dialog should naturally affect the default. It's also repetitive, appearing in so many options right next to each other.

tsujan:
> The preferences dialog should naturally affect the default.

When an option says, for example, "Show main toolbar", its unchecking should remove the toolbar immediately or after closing the Preferences dialog (preferably, the first). But when it says, "Show main toolbar by default", the user knows it changes a default behavior.

tsujan:
I tested it. I'm afraid it seems to me like too many tabs for a few options.

We need a third opinion. @stefonarch, would you please test it and see if you like its Preferences dialog?

xiota:
Maybe Thumbnails could be combined with Image.  Image will become crowded later though.

Or maybe Thumbnails could go with General.

tsujan:
Let's wait for @stefonarch!

tsujan (approved):
This is much better. Thanks!

xiota:
Glad you like this better.  Let me know when you feel like looking at any other PRs.  Also, what's your opinion of using Exiv2 instead of libexif or as a compile-time selection?  Reason I'm thinking about it is Exiv2 supports more file types.

stefonarch:
> Let's wait for @stefonarch!
> 
> EDIT: I asked @stefonarch's opinion on 5 tabs; didn't notice that they're 4 now.

Sorry if I missed something, but it isn't easy to track so many GH notifications. Preferences look good now.

tsujan:
> but it isn't easy to track so many GH notifications.

I use Thunderbird and filter LXQt mails, so that they go to an LXQt folder.

stefonarch:
> > but it isn't easy to track so many GH notifications.
> 
> I use Thunderbird and filter LXQt mails, so that they go to an LXQt folder.

I do so too, but when there are >30 in the morning I prefer to overfly the conversation on GH and  looks I missed something.  I thought about an additional filter for @stefonarch ...

tsujan:
> ... but when there are >30 in the morning...

It happens here too. Most of them are Weblate translation notifications — "Translations update from Weblate" as a filter might make it better for you.

stefonarch:
I've a filter for them too, they doesn't get mixed witnthe rest. But last days just _lxqt_ was quite busy :)

xiota:
@stefonarch

> But last days just lxqt was quite busy :)

If that was from me, I will slow down soon.

tsujan:
> If that was from me,

No one needs to worry about notifications. Of course, it's good practice not to write everything that comes to one's mind as a comment but your comments haven't been of that sort.

Concise comments are always preferable. Rarely, some people write short stories, instead of reports.

PR #448, opened 2021-09-04
Fixed wheel scrolling on image with touchpad

tsujan:
Fixes https://github.com/lxqt/lximage-qt/issues/444

Supercedes https://github.com/lxqt/lximage-qt/pull/446


xiota:
This makes scrolling a lot better.  It's still possible to scroll past dozens of photos at a time.  I can see how that might be desirable when thumbnails are visible, but when they're not, it still looks like it's arbitrarily skipping photos.    Maybe wait for the image to load when thumbnails aren't visible?

    if (loadJob_ && !thumbnailsDock_)
      break;

tsujan:
> Maybe wait for the image to load

Image loading has nothing to do with it.

> It's still possible to scroll past dozens of photos at a time

As it is with mouse.

tsujan:
> Maybe accumulating delta in libfm-qt would help with the other scroll issue I'm having?

No. Deltas are *already* accumulated by `libfm-qt` in the icon/thumbnail mode, with mouse and touchpad alike.

tsujan:
I merge this because it makes touchpad like mouse. I can scroll more images with a sudden movement of the mouse wheel than by a sudden two-finger touchpad scrolling. The rest depends on user's touchpad settings.

PR #450, opened 2021-09-04
Allow creation of shortcut to rename image directly

xiota:
Currently, the only way to create a shortcut to rename a file is to use the File Properties dialog.  This change allows direct access to inline renaming when the thumbnail bar is visible.  When thumbnails are not visible, the libfm-qt rename dialog is shown.


tsujan (approved):
If it wasn't elegant, I'd say it's a duplication because the renaming can already be done by an existing dialog (properties) and, after all, it isn't the job of an image viewer. But good work!

xiota:
Thanks.

PR #454, opened 2021-09-06
Add option to show/hide menubar

xiota:
This is just the final feature needed for a "minimal" interface (#283 and #284).  There is no option in the menubar to hide the menubar because hiding it that way may confuse users who don't know how to restore it.  There is a context menu option and default shortcut Ctrl+M so that the menu may be restored when it is hidden.

I also assigned a default shortcut for renaming, since I had overlooked that earlier.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -121,6 +121,33 @@ MainWindow::MainWindow():
   setShowThumbnails(settings.showThumbnails());
   ui.actionShowThumbnails->setChecked(settings.showThumbnails());
 
+  addAction(ui.actionMenubar);
+  ui.menubar->setVisible(settings.isMenubarShown());
+  ui.actionMenubar->setChecked(settings.isMenubarShown());
+  connect(ui.actionMenubar, &QAction::triggered, this, [this](int checked) {
+    if(!isFullScreen()) { // menubar is hidden in fullscreen
+      ui.menubar->setVisible(checked);
+    }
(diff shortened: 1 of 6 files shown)


tsujan:
Please make these changes:

 1. In Preferences, change "Show menubar" to "Show menubar by default" (don't avoid "by default"; it's needed).
 2. Add the new menubar action to the View menu, immediately before "Main Toolbar".

src/mainwindow.cpp
```diff
@@ -121,6 +121,33 @@ MainWindow::MainWindow():
   setShowThumbnails(settings.showThumbnails());
   ui.actionShowThumbnails->setChecked(settings.showThumbnails());
 
+  addAction(ui.actionMenubar);
+  ui.menubar->setVisible(settings.isMenubarShown());
+  ui.actionMenubar->setChecked(settings.isMenubarShown());
+  connect(ui.actionMenubar, &QAction::triggered, this, [this](int checked) {
+    if(!isFullScreen()) { // toolbar is hidden in fullscreen
```
tsujan, line 128: You mean "*menubar* is hidden in fullscreen".

xiota: yes... copy/paste... will fix it.

xiota:
Forgot to add "by default" back after the GUI change because this was in a different branch.  I've fixed the comment and put the option to hide the menubar in the menu.

tsujan (approved):
Thanks!

PR #455, opened 2021-09-06
Add single window mode

xiota:
Borrowed from pcmanfm-qt "Single window mode".  Only one window and one image will be opened at a time.  A prominent warning is displayed so users will be aware of what happens when they use this option.

screenshot-single-window

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -202,13 +202,44 @@ MainWindow* Application::createWindow(bool fullscreen) {
 
 void Application::newWindow(QStringList files, bool fullscreen) {
   LxImage::MainWindow* window;
-  if(files.empty()) {
+
+  if(settings_.singleWindowMode()) {
+    MainWindow* window = MainWindow::lastActive();
+    // if there is no last active window, find the last created window
+    if(window == nullptr) {
+      QWidgetList windows = topLevelWidgets();
+      for(int i = 0; i < windows.size(); ++i) {
+        auto win = windows.at(windows.size() - 1 - i);
+        if(win->inherits("LxImage::MainWindow")) {
+          window = static_cast<MainWindow*>(win);
+          break;
+        }
+      }
+    }
+    if(window == nullptr) {
+      window = createWindow(fullscreen);
+      window->resize(settings_.windowWidth(), settings_.windowHeight());
+      if(settings_.windowMaximized()) {
+        window->setWindowState(window->windowState() | Qt::WindowMaximized);
+      }
+      window->showAndRaise();
+    }
+    if(!files.empty()) {
+      window->openImageFile(files.first());
+      window->resize(settings_.windowWidth(), settings_.windowHeight());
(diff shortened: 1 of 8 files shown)


tsujan:
Nice!

I'll review it later but please change "Reuse existing window" to "Single window mode" (it may need a concise tooltip, telling the user what he should expect). Also, the Window tab may be more appropriate for this option.

xiota:
I've made the changes you've suggested.

tsujan:
The code seems good but there're a usability problem.

When the user opens multiple images at once in the single-window mode, only the first image is opened. That was why I wasn't fond of such an option in an image viewer.

Since we don't want to add tabs just for this use case, the only solution that comes to my mind is opening the first image in the current window and the others in separate windows. Also, adding "as far as possible" to the tooltip of the new option seems like a good idea to me.

xiota:
I'll think about it.  But the point of having a single window mode is that it not open new windows.  Users who want new windows to open can just not check the box, which is unchecked by default.  Users can still explicitly open new windows if they want.

tsujan:
It's a bug that should be fixed.

xiota:
What is a bug?

stefonarch:
I like much a single window mode, it's meaning is clear. 
No idea if just a message could be shown in the case when multiple images are to be opened.

xiota:
@tsujan  @stefonarch   I noticed that the preferences window has a label for messages.  When  single window mode is clicked/unclicked, it could be used to let the user know what will happen when multiple images are opened.  I think the messages shown there are red and bold, so hard to miss.

When multiple images are opened, they could be shown in the thumbnail dock as a "collection".  Users could create a slideshow limited to just a few images.   I don't know how to make folderview do that though, if it's possible.

tsujan (changes requested):
Depriving the user of an elementary feature like opening multiple images isn't a good idea. We can't compare lximage-qt with pcmanfm-qt in this regard because the latter has tabs.

There can be two solutions, at least. I mentioned one above but it's more complex than my comment. Actually, it isn't a single-window mode anymore but about reusing existing windows. It can confuse the user too.

The second solution is doing what a single-window video player does: it enqueues files when there is more than one. The app needs to show the first image and also a list dialog that contains the names of all images, so that the user could see the other images by clicking its rows (images may be inside different folders).

xiota:
Adding this as an *option* does not deprive anyone of anything.  The program will continue to open new windows when the box is not checked.  However, not adding this option or making it open multiple windows, even though it is in *single* window mode, does deprive users of a single-window mode that is true to its name.

One reason for these PRs is to add features I consider essential.  While a couple PRs were not for essential features (hiding the menubar), I do consider a *real* single window mode to be essential.  So I will not agree to changes that involve making it open new windows without explicit user action via the "new window" command.

I have no objection to a queuing method that does *not* open new windows ("a list dialog").  However, I would leave them for a later PR.

PR #457, opened 2021-09-10
Update lximage-qt_uk.ts

TarasPanchenko:
Виправлення помилки у вікні "Про програму"
Fix error in "About" window


yan12125:
Hi, thanks for the fix. Generally translations should be done on https://translate.lxqt-project.org/. Some translation components were locked due to git rebase failures, and I've fixed all of them. You can continue the awesome work there :)

PR #459, opened 2021-09-13
Open multiple images when dropped

xiota:
Looks like multi-window mode was broken in the first place.  Here's a fix if you want it, unless it's supposed to only open one file when multiple files are dropped.  If that's the case, I have no idea what you were complaining about with the single-window mode PR.

### DIFF
--- a/src/graphicsscene.cpp
+++ b/src/graphicsscene.cpp
@@ -21,6 +21,7 @@
 #include "graphicsscene.h"
 #include <QMimeData>
 #include <QUrl>
+#include <QDir>
 
 namespace LxImage {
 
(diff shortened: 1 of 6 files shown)


tsujan:
> unless it's supposed to only open one file when multiple files are dropped

Yes, it is. We drop files into a single window but can open files by clicking or using command-line.

> I have no idea what you were complaining about with the single-window mode PR.

I really don't know how clearer I could have said it. Since lximage-qt cannot open multiple files in a single window, a single-window mode doesn't make much sense for it. If you want to force it, you'll need to compensate for its problems in some way.

In contrast, pcmanfm-qt, FeatherPad,... can open multiple files in a single window. Therefore, (1) they can open multiple files with DND inside a single window and (2) the concept of "single-window mode" makes sense for them.

xiota:
> > unless it's supposed to only open one file when multiple files are dropped
>
>Yes, it is.

Okay.  So looks like...

> ... there're a usability problem.
>
> When the user opens multiple images at once... only the first image is opened.

PR #464, opened 2021-09-21
Fixed issues in a recent commit about thumbnails positions

tsujan:
Excluded right area (because it's used by EXIF data) and fixed localization of position strings.

@stefonarch, please merge it if your tests show it's OK. After that, you could update translations; my next PR won't add a translatable string.


stefonarch (approved):
GTM

stefonarch:
Translations update will follow later. 
Btw I found another issue (context menu in free space in thumbnail bar has some not working items) will open an issue later.

tsujan:
> context menu in free space in thumbnail bar has some not working items

It works here but, of course, changing of sorting or showing of hidden files is temporary, as it should be.

Issue #470, opened 2021-09-25
Horizontal flipping is broken


Dialga:
##### Expected Behavior

Press H to horizontally flip image.
##### Current Behavior

Pressing H to horizontally flip image flips the image horizontally then vertically.
##### Possible Solution

Press H to horizontally flip image only.
##### Steps to Reproduce (for bugs)

1. Open jpg image
2. Press H
3. Image is flipped horizontally
4. Image is flipped vertically

##### System Information

* LXimage-qt version: 0.17.0
* Distribution & Version: Arch Linux
* Qt Version: 5
* liblxqt Version: 0.17.0

Dialga:
Not sure if this is related, but there is a duplicate line here:

https://github.com/lxqt/lximage-qt/blob/b771fe7efe633419f10a749166a625cb4e5b773c/src/mainwindow.cpp#L205-L206

tsujan:
Thanks for the report! I'll fix the problem soon.

tsujan:
The fix is here: https://github.com/lxqt/lximage-qt/pull/471

It was a single typo that, by chance, had effect only when the image was made smaller than its original size and also when it wasn't SVG or GIF. That was why the bug had remained hidden.

The redundant line that you mentioned was a harmless typo; I removed it in the same PR.

tsujan:
BTW, thanks to this report, I found another bug in flipping (whether vertical or horizontal), for which I'll open an issue and add a fix later.

Dialga:
Glad I can be of help!

> BTW, thanks to this report, I found another bug in flipping (whether vertical or horizontal), for which I'll open an issue and add a fix later.

tsujan:
> Glad I can be of help!

A great help, indeed! The issue I've found is nasty and, if you didn't make this report, it might go into the next version (in November).

PR #471, opened 2021-09-25
Fixed horizontal flipping of resized images

tsujan:
It was caused by a typo in the second argument of `QImage::mirrored` and made images also flip vertically when they were made smaller than their original sizes.

Fixes https://github.com/lxqt/lximage-qt/issues/470


tsujan:
There's more work to do (see https://github.com/lxqt/lximage-qt/issues/472). Merging.

Issue #473, opened 2021-09-25
[Feature request] Show version number in about window


Dialga:
**Is your feature request related to a problem? Please describe.**

Currently clicking on the about dialog, there is no version number of lximage.

**Describe the solution you'd like**

The about window includes the version number of lximage.
The about dialog in applications normally show the version number of the application that's running, that way an end user can quickly report about buggy versions.

**Describe eventual alternatives you've considered**

#### Context

Arch Linux

stefonarch:
True, pcmanfm-qt and screengrab report their versions.

Dialga:
> But actually the copyright should also be updated.

Yes I was also going to mention that.

tsujan:
In addition, why limiting it to plain-text? Will make it like this after https://github.com/lxqt/lximage-qt/pull/474 is merged:

![about](https://user-images.githubusercontent.com/981076/135035646-e6cd47a2-0454-4798-a739-3b44962b7683.png)

It removes unnecessary characters and strings from translation, so that it won't be broken by updating the version and copyright.

stefonarch:
Nice!

stefonarch:
Btw, version number, will we stick to 0.1x  still?

tsujan:
> Btw, version number, will we stick to 0.1x still?

I don't know. Next month, we could ask other members. What's your opinion?

stefonarch:
> I don't know. Next month, we could ask other members. What's your opinion?

I'm fine with everything, but going over 0 would be nice at the point we are. 
I saw today a typo [here](https://emmabuntus.org/on-september-20th-2021-emmade4-under-debian-11-bullseye-stable-version/), LXQt 8.16.0 way far in the future but very nice version number  ;)

tsujan:
> I'm fine with everything, but going over 0 would be nice at the point we are. 1.17?

If we want to bump the first number, it'll be 1.0.0.

> Btw a very well done and complete distribution

Wherever I see an old version of LXQt, all the bugs that were fixed after it come to my mind ;) Also 0.17 will be old soon.

stefonarch:
> Wherever I see an old version of LXQt, all the bugs that were fixed after it come to my mind ;) Also 0.17 will be old soon.

That's natural for who is in, but I'd say from 0.16 most things where quite ok, no need to talk too bad about ;)

PR #478, opened 2021-09-28
Improved About dialog

tsujan:
Closes https://github.com/lxqt/lximage-qt/issues/473

@stefonarch 
The translatable strings are split for keeping them intact after version/copyright changes. Only one translatable string needs to be changed in such cases: "2013-2021"; in most languages, it doesn't need a translation.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -255,11 +255,15 @@ MainWindow::~MainWindow() {
 
 void MainWindow::on_actionAbout_triggered() {
   QMessageBox::about(this, tr("About"),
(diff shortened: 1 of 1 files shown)


stefonarch:
Looks good. In pcmanfm-qt we include also a link to the github page, maybe add it here too? Could be useful.

tsujan:
> In pcmanfm-qt we include also a link to the github page, maybe add it here too? Could be useful.

Yes, it's a good idea. I don't add it here because I have more patches and want to make sure that the git source is identical to mine (to rule out mistakes). Will add it after the other patches are merged.

Issue #475, opened 2021-09-26
Space and backspace doesn't work as shortcuts for next/previous image


stefonarch:
##### Expected Behavior

As user can assign them, they should work
##### Current Behavior

They are shown in "shortcuts" instead of `> <` but are not working.
##### Possible Solution

Make them work (best) or don't make it possible to select them.
##### Steps to Reproduce (for bugs)

1.Assign space for next and backspace for previous image in Preferences > Shortcuts
2. Close and open lximage-qt
3. try using them
4. only  `> <`  working

##### Context

Those shortcuts are often used to  go through files, examples are darktable, digikam
##### System Information

* LXimage-qt version: git, all

tsujan:
Thanks!

Space, Backspace and a few other shortcuts are hard-coded (Space → next image, Backspace → previous image,...). When they're defined explicitly by the user, I guess they become ambiguous.

I'll look into them later. For now, I'm testing a vast code cleanup. Separate commits were added in previous months, without considering their relations to each other, and that has cluttered the code (and also introduced some bugs).

tsujan:
This can be fixed by defining/updating hard-coded shortcuts after custom shortcuts are set and not duplicating the latter. I wrote the needed code and it worked.

Since I have a considerable amount of changes that aren't committed yet, the order of committing will be this:

 1. Merging https://github.com/lxqt/lximage-qt/pull/474
 2. Making a PR for https://github.com/lxqt/lximage-qt/issues/473 and merging it.
 3. Making a PR for https://github.com/lxqt/lximage-qt/issues/477 and merging it.
 4. Making a PR for the current issue and merging it.

stefonarch:
They should both also be listed by default.

tsujan:
> They should both also be listed by default.

They can't be listed in the shortcut editor because they aren't assigned to any action — they're hard-coded into the window itself.

Maybe a "Hidden Shortcuts" dialog. But it isn't so important because all shortcuts that the user may need can be assigned by him in the shortcut editor.

tsujan:
Oh, and there are 5 of them, not 2.

PR #482, opened 2021-09-30
Allow custom shortcuts to override hard-coded shortcuts

tsujan:
With this patch, hard-coded shortcuts (Escape → exiting/un-fullscreen, Space and Right → next image, Backspace and Left → previous image) can be overridden by custom shortcuts. Previously, overriding resulted in ambiguity, so that the shortcut was disabled.

Also, two hard-coded shortcuts and a "Hidden Shortcuts" dialog are added. The latter tells the user about overriding too.

In addition, the code is cleaned up with regard to custom shortcuts.

Fixes https://github.com/lxqt/lximage-qt/issues/475

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -55,6 +55,7 @@ set(lximage-qt_UIS
     preferencesdialog.ui
     resizeimagedialog.ui
     screenshotdialog.ui
+    shortcuts.ui
 
     upload/uploaddialog.ui
 )
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -166,37 +166,6 @@ MainWindow* Application::createWindow(bool fullscreen) {
   LxImage::MainWindow* window;
   window = new LxImage::MainWindow();
   window->setShowFullScreen(fullscreen);
-
-  // get default shortcuts from the first window
-  if(defaultShortcuts_.isEmpty()) {
-    const auto actions = window->findChildren<QAction*>();
-    for(const auto& action : actions) {
-      if(action->objectName().isEmpty() || action->text().isEmpty()) {
-        continue;
-      }
-      QKeySequence seq = action->shortcut();
-      ShortcutDescription s;
-      s.displayText = action->text().remove(QLatin1Char('&')); // without mnemonics
-      s.shortcut = seq;
-      defaultShortcuts_.insert(action->objectName(), s);
-    }
-  }
-
-  // apply custom shortcuts to this window
-  QHash<QString, QString> ca = settings_.customShortcutActions();
-  const auto actions = window->findChildren<QAction*>();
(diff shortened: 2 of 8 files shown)


tsujan:
@stefonarch, your tests are needed.

stefonarch:
My 2 cents fot the moment:

* I happen to use alt+space for opening runner, and this prevails here. 
I think` End | Home` are enough  for last|first image.

* "These hard coded shortcuts will be overridden if they are also assigned in Preferences" → "These default shortcuts will be overridden if they are  assigned in Preferences to other actions." ?
*  there shows up a comma and an ellipsis - so it looks like more than one shortcut could be assigned and shown, but I found no way to add it. Comma gives a comma as shortcut, shown shortly as  `,, ...`

stefonarch:
My 2 cents for the moment:

* I happen to use alt+space for opening runner, and this prevails here. 
I think` End | Home` are enough  for last|first image.

* "These hard coded shortcuts will be overridden if they are also assigned in Preferences" → "These default shortcuts will be overridden if they are  assigned in Preferences to other actions." ?
*  there shows up a comma and an ellipsis - so it looks like more than one shortcut could be assigned and shown, but I found no way to add it. Comma gives a comma as shortcut, shown shortly as  `,, ...`

stefonarch:
My 2 cents for the moment:

* I happen to use alt+space for opening runner, and this prevails here. 
I think` End | Home` are enough  for last|first image.

* "These hard coded shortcuts will be overridden if they are also assigned in Preferences" → "These default shortcuts will be overridden if they are  assigned in Preferences to other actions." ?
*  there shows up a comma and an ellipsis - so it looks like more than one shortcut could be assigned and shown, but I found no way to add it. Comma gives a comma as shortcut, shown shortly as  `,, ...`

tsujan:
> I think End | Home are enough for last|first image.

OK.

> if they are assigned in Preferences to other actions

OK, again.

> there shows up a comma and an ellipsis - so it looks like more than one shortcut could be assigned and shown

Yes, but by not allowing multiple shortcuts for the same action, the Qt bug that makes Meta a non-modifier is worked around; see https://github.com/lxqt/lximage-qt/blob/c7ae8137ace298df7a7baba9ceafe3b790b8ea25/src/preferencesdialog.cpp#L35 If Qt's Meta bug is fixed, I'll remove that restriction.

tsujan:
Done, in a separate commit.

Home and End were in the GUI file but I hard coded them. If the user removes them in Preferences, they should still work. Of course, if they're used for other actions, they'll be overridden, like other hard-coded shortcuts.

Please test again — especially with your localization — and tell me if it's OK or not.

stefonarch:
Looks all ok.

tsujan:
Thanks for tests!

PR #484, opened 2021-10-01
Updated Readme.md

stefonarch:
Not sure about:
* inserting an image
* Inserting : "Supported image formats: jpg, png, ecc ecc"


tsujan:
> Inserting : "Supported image formats..."

That depends on what's installed on the system.

stefonarch:
Ok, so not practicable. 
Anything else missing? Eventual picture:

schermata-10-01-22-17

tsujan:
I suggest adding something like  "And other features that can be found when it is actually used" because, for example, it can also open images with other apps that handle image mime types, can save images, or have some of libfm-qt's functionalities. It's hard to list all features of an app.

stefonarch:
> adding something like "And other features that can be found when it is actually used"

Done, and also added the screenshot.

README.md
```diff
@@ -4,10 +4,24 @@
...
+* Thumbnail bar (left, top or bottom); different thumbnail sizes
+* Exif data bar
+* Inline image renaming
+* Custom shortcuts
+* Image annotations (arrow, rectangle, circle, numbers)
+* Recent files
+* Upload images (Imgur, ImgBB)
+* Take screenshots
+
+Even more features that can be found using it. LXImage-qt is maintained by the LXQt project
```
tsujan, line 23: The English has a small problem: "...that can be found using it" → "...that can be found when it is used".

stefonarch: Maybe more than one problem...

`More features can be found when it is used.` ?

tsujan: > More features can be found when it is used.

Yes. It's better.

tsujan: BTW, some people might say it's "LXImage-Qt", not "LXImage-qt".

stefonarch: We are the DE with most use of shift key definitevely :)
Will change it.

tsujan:
It's good . Merge it.

PR #488, opened 2021-10-05
Added option to disable blur filter on zoom

### LINKED ISSUE #386: Option to disable smoothing filter for zoom in/out
There should be an option in the preferences to disable the smoothing/blur filter when zooming in on images.

##### Expected Behavior

Allow the user to disable the filter that adds a blur/smoothing effect when scaling an image.

##### Current Behavior

The smooth/blur filter is always enabled when the scale is not 100%.

##### Possible Solution

Add an option to the Preferences menu.

##### Context

I work with pixel art and always need to zoom in to check individual pixels. The filter that blurs the image means it's impossible to use this image viewer and i need to install a different one.

##### System Information

* Distribution & Version: Ubuntu 20.10
* Kernel: Linux  5.8.0-33-generic
* Qt Version: 5.14.2
* libfm-qt Version: 0.15.1-0ubuntu3
* libqtxdg Version: not installed
* lxqt-build-tools Version: not installed
* Package version: 0.15.0-0ubuntu1

redtide:
Fixes #386

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -49,6 +49,7 @@ ImageView::ImageView(QWidget* parent):
   cursorTimer_(nullptr),
   scaleFactor_(1.0),
   autoZoomFit_(false),
+  smoothOnZoom_(true),
   isSVG(false),
   currentTool(ToolNone),
   nextNumber(1),
@@ -611,6 +612,10 @@ void ImageView::updateOutline() {
 }
 
 void ImageView::paintEvent(QPaintEvent* event) {
+  if (!smoothOnZoom_) {
+    QGraphicsView::paintEvent(event);
+    return;
+  }
   // if the image is scaled and we have a high quality cached image
   if(imageItem_ && scaleFactor_ != 1.0 && !cachedPixmap_.isNull()) {
     // rectangle of the whole image in viewport coordinate
@@ -652,7 +657,7 @@ void ImageView::queueGenerateCache() {
 
   // we don't need to cache the scaled image if its the same as the original image (scale:1.0)
   // no cache for gif animations or SVG images either
-  if(scaleFactor_ == 1.0 || gifMovie_ || isSVG) {
+  if(scaleFactor_ == 1.0 || gifMovie_ || isSVG || !smoothOnZoom_) {
     if(cacheTimer_) {
       cacheTimer_->stop();
       delete cacheTimer_;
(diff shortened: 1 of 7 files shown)


src/preferencesdialog.ui
```diff
@@ -201,6 +201,13 @@
          </property>
         </widget>
        </item>
+       <item row="7" column="0">
+        <widget class="QCheckBox" name="useBlurOnZoomBox">
+         <property name="text">
+          <string>Smooth images on zoom</string>
```
tsujan, line 207: You've chosen a good name for the option ("Smooth images on zooming" — an "ing" was missing) because it isn't about blurring. Please change the key, and method names accordingly (e.g., `SmoothOnZoom` instead of `UseBlurOnZoom`, `smoothOnZoom()` instead of `useBlurOnZoom()`, etc.)

tsujan:
Please also add the boolean for disabling smoothing in these lines:

https://github.com/lxqt/lximage-qt/blob/806bfc1315b630b74f3cdb05c2c8b7003d2c5e4c/src/imageview.cpp#L655

https://github.com/lxqt/lximage-qt/blob/806bfc1315b630b74f3cdb05c2c8b7003d2c5e4c/src/imageview.cpp#L680

src/imageview.cpp
```diff
@@ -611,6 +612,10 @@ void ImageView::updateOutline() {
 }
 
 void ImageView::paintEvent(QPaintEvent* event) {
+  if (!useBlurOnZoom_ && scaleFactor_ != 1.0) {
```
tsujan, line 615: `scaleFactor_ != 1.0)` isn't needed here.

redtide:
Done

src/preferencesdialog.ui
```diff
@@ -201,6 +201,13 @@
          </property>
         </widget>
        </item>
+       <item row="7" column="0">
+        <widget class="QCheckBox" name="smoothOnZoomBox">
+         <property name="text">
+          <string>Smoothing images on zoom</string>
```
tsujan, line 207: Please just change this to `Smooth images on zooming`. If it's changed after the merge, translations might be removed.

redtide: oops, sorry, done

tsujan (approved):
GTM. Thank you very much!

Issue #490, opened 2021-10-08
Shortcuts may not work without menubar.


tsujan:
##### Current Behavior

Several shortcuts don't work when the menubar isn't shown by default and the app is restarted.

##### Steps to Reproduce (for bugs)

1. Hide menubar by default.
2. Restart the app.
3. Try to use the shortcut for showing thumbnail bar.

It doesn't work.

##### Context

I should have reviewed some PRs more carefully. Now, I have to find and fix their problems :(

##### System Information

* Latest git.

PR #492, opened 2021-10-08
Fixed lost shortcuts when app starts without menubar

tsujan:
Fixes https://github.com/lxqt/lximage-qt/issues/490

@stefonarch 
As usual, your check is needed before merging.


stefonarch (approved):
GTM, found none not working now, first P=preferences didn't work here.

PR #517, opened 2022-01-01
Spelling: Exif

comradekingu:
Unfortunately https://en.wikipedia.org/wiki/Exif
Consistent with https://translate.lxqt-project.org/translate/lxqt-desktop/lximage-qt/en/?checksum=77cd9fbe7eaa5cbe (Show Exif data dock by default)

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -554,7 +554,7 @@
     <bool>true</bool>
    </property>
    <property name="text">
-    <string>Show EXIF Data</string>
+    <string>Show Exif Data</string>
    </property>
   </action>
   <action name="actionDrawNone">


tsujan:
Exif or EXIF, it doesn't make a important difference → https://github.com/libexif/libexif#readme

comradekingu:
@tsujan 
It is important, and it is officially _Exif_, according to JEIDA/JEITA/CIPA specifications.
The library you linked to has it wrong.
Doing both "Exif" and "EXIF" is as bad as it gets, and that is the mistake you are upholding.

I checked what the correct way to do it is, and I put in a fix for it.
I'll happily try to correct it in libexif if that is the source you want to go by.

Issue #519, opened 2022-01-07
"Upload" should be in "File" menu


stefonarch:
and not under "Modify".

Not a big thing, just cleaner. 
                                      
* LXimage-qt version :git, 1.0

PR #521, opened 2022-01-07
Move upload entry, add shortcut

stefonarch:
Fixes https://github.com/lxqt/lximage-qt/issues/519
Added also a shortcut.


tsujan (approved):
GTM.

In my own projects, I wouldn't add a new shortcut because it might be in conflict with a custom shortcut that's added by the user. But it isn't a big deal.

PR #523, opened 2022-01-11
Removed Screenshot Desktop file

slidinghotdog:
Removed since it's deprecated.


tsujan (approved):
GTM.

Any objection to merging it? The screenshot code has more than one problem and we'd better not encourage users to use it instead of screengrab.

stefonarch:
GTM, but all the lximage-qt-screenshot*.desktop.yaml files should also be removed.

tsujan:
@stefonarch, would you remove them?

PR #530, opened 2022-01-29
Restore maximized state (if any) on un-fullscreening

tsujan:
The best method of fullscreening is check the maximized state before fullscreening and restore it on un-fullscreening. Previously, `showFullScreen()` was used; it removed the max state before fullscreening.

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -1177,11 +1177,8 @@ void MainWindow::on_actionPrint_triggered() {
 }
 
 // TODO: This can later be used for doing slide show
-void MainWindow::on_actionFullScreen_triggered(bool checked) {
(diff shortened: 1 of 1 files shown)


ralf3u:
With F11 & F11 (do/undo) the issue does not appear, but with View>Full Screen (menu-bar) & F11, sometimes with F11 & Esc, and also sometimes with View>Full Screen (menu-bar) & Esc.

By the way, in my lximage-qt (LXQt 0.17.0) there is a checkbox in the menu-bar next to Full Screen in View. I would delete the checkbox because it makes no sense for me.

tsujan:
@ralf3u 
Please don't add irrelevant comments to PRs.

Issue #557, opened 2022-06-08
Top of image gets corrupted when toolbar extension popup slides back


ArrayBolt3:
##### Expected Behavior
When the LXImage window is somewhat small, there's a button at the far right of the toolbar that make a drop down appear where the overflowed buttons are located. When this drop down slides back up, the image should cleanly reappear where the drop down was covering it.

##### Current Behavior
The part of the image that got covered by the drop down gets pretty badly mangled as the drop down slides back up. The top of the image ends up getting overwritten by what appears to be multiple highly vertically compressed copies of the image stacked on top of each other. This only happens with certain window sizes (though it's quite easy to replicate), and only happens if the drop down covers part of the image. It also vanishes fairly quickly if there's enough blank window area above the image, whereas it appears to stick if the image top is close to the image area border.

##### Possible Solution
Looks like the thing is writing the whole entire image every redraw, but only writing it to the redrawn area. If this is the case, it might be possible to shrink the image as the drop down descends, and then grow it as it slides up, or make the redraw only happen once when the drop down is fully ascended. Or maybe do it the hard way by using complex math tricks to figure out what portion of the image needs redrawn and only just redraw that, so that the drop down appears to uncover the image. I'd guess this last one is what LXImage tries to do based on its behavior with certain window sizes, but that there's a bug somewhere in it. This is all just my guesswork - I've not actually looked at the code.

##### Steps to Reproduce (for bugs)
1. Open the Application Menu, hover over Graphics, and click LXImage.
2. Click "File", then "Capture Screenshot".
3. Ensure that you're capturing the whole screen with a delay of 0 seconds, then click "OK".
4. Grab the lower-right corner of the LXImage window, and shrink it somewhat. Ensure the top of the image is touching the top of the image viewing boundary, and that the right-most button of the toolbar looks like a right angle bracket.
5. Click the right angle bracket button to cause the drop down to descend, then click it again to slide it up. Bam, image goes glitchy.

##### Context
Glitchy image = weird, annoying, wrong. Non-glitchy image = yay.

##### System Information
* LXimage-qt version: 1.1.0
* Distribution & Version: Lubuntu 22.10 Alpha
* Qt Version: 5.15.4
* liblxqt Version: 1.1.0-0ubuntu2

tsujan:
This video shows the problem:

https://user-images.githubusercontent.com/981076/172735448-70cd7bb2-e962-47c4-ad74-dba37f03bf3b.mp4

[...]

PR #558, opened 2022-06-09
Fixed repainting of cached images

tsujan:
Previously, the cached pixmap was drawn inside the repaint rectangle. That was wrong because there's no guarantee that the repaint rectangle is as big as the cached pixmap (e.g., it isn't so when the toolbar extension popup covers a part of the image and slides back). Therefore, visual glitches were imminent in special cases.

This patch draws the cached pixmap inside its corresponding cache rectangle, so that the image is always repainted correctly. The paint clipping is done by Qt automatically, as QPaintEvent's doc explains.

Fixes https://github.com/lxqt/lximage-qt/issues/557


tsujan:
Any objection to merging?

Issue #592, opened 2023-01-06
PCX not supported by QImageReader, should not be listed in .desktop file


fenugrec:
https://github.com/lxqt/lximage-qt/blob/3bc6161c23b490737d340943e98bdbcd108ed2e4/data/lximage-qt.desktop.in#L8

`application/pcx` and `image/x-pcx` should be removed from `data/lximage-qt.desktop.in`, unless I'm misunderstanding something or missing a non-obvious optional dependency ?

At least, PCX is not listed in the qt6 docs for  qimagereader :
https://doc.qt.io/qt-6/qimagereader.html#formats

For reference, the attached image cannot be opened, as expected.
[S5.PCX.zip](https://github.com/lxqt/lximage-qt/files/10363329/S5.PCX.zip)

tsujan:
"pcx" is among the formats supported by Qt5, on which lximage-qt is based currently.

But you're right: (1) Qt5 can't handle it (correctly), and (2) Qt6 doesn't support it. It will be removed when lximage-qt is ported to Qt6.

tsujan:
BTW, a funny comparison:

Qt5 supports "3fr", "ani", "arq", "arw", "avif", "avifs", "bay", "bmp", "bmq", "bw", "cap", "cine", "cr2", "cr3", "crw", "cs1", "cur", "dc2", "dcr", "dcs", "dng", "drf", "dxo", "eip", "eps", "epsf", "epsi", "erf", "exr", "fff", "gif", "hdr", "heic", "heif", "icns", "ico", "iiq", "jp2", "jpeg", "jpg", "jxl", "k25", "kc2", "kdc", "kra", "mdc", "mef", "mfw", "mng", "mos", "mrw", "nef", "nrw", "obm", "ora", "orf", "ori", "pbm", "pcx", "pdd", "pdf", "pef", "pgm", "pic", "png", "ppm", "psb", "psd", "psdt", "ptx", "pxn", "qtk", "r3d", "raf", "ras", "raw", "rdc", "rgb", "rgba", "rw2", "rwl", "rwz", "sgi", "sr2", "srf", "srw", "sti", "svg", "svgz", "tga", "tif", "tiff", "wbmp", "webp", "x3f", "xbm", "xcf", "xpm".

[...]

PR #595, opened 2023-01-13
Remove image/tiff from lximage-qt.desktop

alexhenrie:
LXImage-Qt can only open TIFF files if the qt5-imageformats module is installed.

Per the discussion at <https://github.com/lxqt/lximage-qt/issues/592#issuecomment-1374221733>, the .desktop file should only advertise support for formats that LXImage-Qt can open *without* installing the qt5-imageformats module.

### DIFF
--- a/data/lximage-qt.desktop.in
+++ b/data/lximage-qt.desktop.in
@@ -5,4 +5,4 @@ Exec=lximage-qt %F
 StartupNotify=true
 Terminal=false
 Categories=Graphics;Viewer;RasterGraphics;2DGraphics;Photography;
-MimeType=image/bmp;image/gif;image/jpeg;image/jpg;image/png;image/tiff;image/x-bmp;image/x-pcx;image/x-tga;image/x-portable-pixmap;image/x-portable-bitmap;image/x-targa;image/x-portable-greymap;application/pcx;image/svg+xml;image/svg-xml;
+MimeType=image/bmp;image/gif;image/jpeg;image/jpg;image/png;image/x-bmp;image/x-pcx;image/x-tga;image/x-portable-pixmap;image/x-portable-bitmap;image/x-targa;image/x-portable-greymap;application/pcx;image/svg+xml;image/svg-xml;


alexhenrie:
Actually, I may have misunderstood the discussion. Is the the idea to only advertise support for formats that Qt supports without plugins, or is the idea to advertise everything that qt5-imageformats and kimageformats support? If the idea was to include everything then `image/tiff` should be kept and other formats like `image/webp` should be added too.

tsujan:
> Per the discussion at https://github.com/lxqt/lximage-qt/issues/592#issuecomment-1374221733,...

Thanks for caring, but please read that page to the end. Also, see [the commit](https://github.com/lxqt/lximage-qt/commit/435946e45dae81a265aa8ca33f6293792db0c1fe) that closed it.

tsujan:
> and other formats like image/webp should be added too.

We don't want to overdo it. Users can add associations.

tsujan:
We've add a few formats that aren't supported by Qt just to tell users that they might want to install a format plugin, but not more. File association exists for more.

As for `image/webp`, I agree that it has become a common format. You could make a PR for it.

alexhenrie:
Thank you for the answer. I will submit a pull request for `image/webp`. While we're on the subject, what about `image/vnd.microsoft.icon` (.ico) files, which nearly every program that is installed in Wine includes, and are also used as the "favicons" of web pages? The trouble is that "common" is a relative/subjective standard, so it's not clear to me whether ICO is as "common" as WebP.

tsujan:
> I will submit a pull request for image/webp

Thanks for your contribution.

> what about image/vnd.microsoft.icon (.ico) files

I really think it would be an overkill, especially because it isn't native. This is one of cases where we should resist temptations; otherwise, it would have no end ;)

alexhenrie:
> > what about image/vnd.microsoft.icon (.ico) files
> 
> I really think it would be an overkill, especially because it isn't native.

Then shouldn't we drop image/x-targa, which also requires qt5-imageformats and isn't exactly "native" either?

tsujan:
Default settings are just what an app sets by default. They can't be what all or most users may want. That's why apps and DEs are customable. Hence talking about an end.

PR #596, opened 2023-01-13
Add image/webp to lximage-qt.desktop

alexhenrie:
And alphabetize the MIME type list to make it easy to tell what's included and what's not.


tsujan (approved):
Thanks!

PR #608, opened 2023-04-04
Correct duplicate names for two QFormLayout in preferencesdialog.ui

isf63:
Corrects two QFormLayout with identical name "formLayout_1".

This suppresses compile time warning:
AutoUic: .../lximage-qt-git/src/lximage-qt/src/preferencesdialog.ui: Warning: The name 'formLayout_1' (QFormLayout) is already in use, defaulting to 'formLayout_11'.


tsujan (approved):
Thanks!

PR #621, opened 2023-05-23
Port away from QPainter::HighQualityAntialiasing

luis-pereira:
QPainter::HighQualityAntialiasing and will be ignored. Just use the
Antialiasing render hint.

### DIFF
--- a/src/mainwindow.ui
+++ b/src/mainwindow.ui
@@ -37,7 +37,7 @@
     <item>
      <widget class="LxImage::ImageView" name="view">
       <property name="renderHints">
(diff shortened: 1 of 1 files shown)


tsujan:
Already done in https://github.com/lxqt/lximage-qt/pull/600.

Please also see https://github.com/lxqt/libfm-qt/issues/889, which is relevant here too (5.15 being the last version of Qt5).

tsujan:
> The goal of this commit was to port the deprecated stuff before the Qt6 port.

Yes, I know. Removal of deprecated stuff is very good for apps that haven't been ported yet (lxqt-panel, lxqt-config,....); it'll make porting much easier.

PR #626, opened 2023-07-10
Replace instances of 'LXImage' with 'LXImage-Qt' in desktop entries

isf63:
Also fix capitalization in several places


stefonarch:
LGTM

PR #628, opened 2023-07-13
Follow freedesktop spec for .desktop file and metainfo.xml

stefonarch:


palinek:
I'm just curious, why is there the replacement of dash to underscore in lxqt-project, but not in LXimage-qt?

stefonarch:
This is still to decide yet, IMO: All underscores, only in domain, no underscore. Atm I liked that more as it's  a "well-known" name.

>Well-known names containing the dash are allowed but not recommended, because the dash is not allowed in some related uses of reversed DNS names, such as D-Bus object paths and interface names, and Flatpak app IDs. If the author's domain name contains a dash, replacing it with an underscore is recommended: this cannot cause ambiguity, because underscores are not allowed in DNS domain names.

redtide:
Maybe this is more clear:
<https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-id-generic>

> To ensures the greatest possible compatibility of an AppStream ID, it is recommended to replace any hyphens in the ID with underscores, and prefix every leading digit of a section with an underscore as well. Since the underscore is not a valid character in domain names, the uniqueness of the ID is kept. For example, the ID org.7-zip.7zip could become org._7_zip._7zip.

This means underscore everywhere, also the name (and so as prefix if also the name has a digit, not just the domain)

redtide:
@palinek maybe it might be interesting that [string(MAKE_C_IDENTIFIER \<string\> \<output_variable\>)](https://cmake.org/cmake/help/latest/command/string.html#make-c-identifier) seems to do what desktop entry spec and appstream wants:

> Convert each non-alphanumeric character in the input \<string\> to an underscore and store the result in the \<output_variable\>. If the first character of the \<string\> is a digit, an underscore will also be prepended to the result.

[here](https://github.com/qtilities/sqeleton/blob/master/cmake/AppStream.cmake#L13) my attempt to deal with the rules on CMake,

Issue #320, opened 2020-05-13
[Enhancement] Color management support


nvmnghia:
Add support for color management.

##### Expected Behavior

Support both embedded and system monitor profile. Given that the whole Lubuntu environment doesn't support color management out of the box, I suggest that LXImage should at least support the embedded color profile.

This [displaycal](http://displaycal.net/icc-color-management-test/) test was used to determine color profile support.

##### Current Behavior

It doesn't support at all, please follow the above link.

##### Possible Solution

I don't know :p

##### Context

I take a lot of photos in DNG, and output to sRGB and DCI-P3 (sometimes AdobeRGB).

##### System Information

* Distribution & Version: Lubuntu 20.04

tsujan:
I have no experience in color management but I know that Qt can get embedded ICC profiles. However, I didn't find a Qt method for applying them.

gwenview has it own codes for it but, as far as I tested, the results were suboptimal.

tsujan:
This report shows that it isn't supported yet: https://bugreports.qt.io/browse/QTBUG-6582

HybridDog:
The main purpose of an image viewer is to display images, so in my personal opinion, color management is a necessary requirement for me to use the software.
In comparison to firefox, eye of gnome and mpv (with `--vo=gpu-next`), lximage-qt shows this image in green:
![image](https://github.com/lxqt/lximage-qt/assets/3192173/933b79a9-eef1-456d-8062-722c5d04618b)

Another necessary requirement for me is [gamma-correct scaling](http://www.ericbrasseur.org/gamma.html?i=2)

Here is a website which may have useful information about color management: https://gitlab.freedesktop.org/pq/color-and-hdr/-/tree/main

> This report shows that it isn't supported yet: https://bugreports.qt.io/browse/QTBUG-6582

The last comment mentions that ICC profiles are supported except the parsing of "printer style color profiles".

tsujan:
The report is open, and we accept PRs.

tsujan:
Not having dealt with color spaces before, I just added `QImage::convertToColorSpace(QColorSpace(QColorSpace::SRgb))` to lximage-qt's code and tested it with a few images which had embedded color profiles, including the image of the previous comment. The result seemed fine (compare the displayed image with its thumbnail):

![embedded](https://github.com/lxqt/lximage-qt/assets/981076/9e27df40-8562-4aa0-9c93-48d75e964a75)

[...]

PR #635, opened 2023-09-11
Added minimal support for color spaces

tsujan:
… as far as Qt allowed.

Closes https://github.com/lxqt/lximage-qt/issues/320


tsujan:
See the report for the performed tests. Merging...

PR #666, opened 2024-04-25
Fixed a crash on using annotations with Qt ≥ 6.7

tsujan:
This is similar to https://github.com/lxqt/lxqt-notificationd/pull/375. Both fixes are logical, regardless of what has changed in Qt 6.7.


stefonarch (changes requested):
Works again, GTM

stefonarch (approved):

PR #677, opened 2024-06-18
Small fix to screenshot select area

isf63:
Set QFrame::NoFrame on ScreenshotSelectAreaGraphicsView to prevent the main view from being oversized. Previously it's possible to use the arrow keys to shift the view 1px.

Additionally, another change should be made for allowing to select the entire screen area (it's off my 1px regardless of this.)


tsujan (approved):
GTM. Thanks.

Issue #690, opened 2024-12-08
Cannot open very large images


avgoor:
In the recent version of the package there is an issue with loading large images, seems due to QT defaults.
```
$ file ./4902762196_34f3877ea8_o.jpg
./4902762196_34f3877ea8_o.jpg: JPEG image data, JFIF standard 1.02, resolution (DPI), density 300x300, segment length 16, Exif Standard: [TIFF image data, little-endian, direntries=15, width=10538, height=7310, bps=194, compression=none, PhotometricInterpretation=RGB, description=II*, orientation=upper-left], baseline, precision 8, 10538x7310, components 3

$ lximage-qt ./4902762196_34f3877ea8_o.jpg 
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
qt.gui.imageio: QImageIOHandler: Rejecting image as it exceeds the current allocation limit of 256 megabytes
```
##### Expected Behavior
Such a large image is displayed

##### Current Behavior
Empty window and `qt.gui.imageio: QImageIOHandler: Rejecting image as it exceeds the current allocation limit of 256 megabytes` in the terminal output.

##### Possible Solution
It seems the possible solution could be setting the limit explicitly:
[https://doc.qt.io/qt-6/qimagereader.html#setAllocationLimit](https://doc.qt.io/qt-6/qimagereader.html#setAllocationLimit)

##### Steps to Reproduce (for bugs)
1. Open an 10000x10000 JPEG image 

##### Context
Just inconvenient to use the app when scrolling the folder containing large images

##### System Information
* LXimage-qt version: 2.1.0
* Distribution & Version: ArchLinux 
* Qt Version: 6.8.1-1
* liblxqt Version: none

tsujan:
Interesting! Are you saying that such images were shown with Qt5 without problem? Because, if that was the case, setting `mbLimit` to zero may be an option.

> Open an 10000x10000 JPEG image

If you have one, please attach it. The largest image I have for testing is 8736×5856. It's an HEIF image and is opened without problem.

avgoor:
Yes, I had no problems on Qt5. When ArchLinux moved to Qt6 I noticed this issue. I'm attaching the same image that I have problems with. Hope this helps.

[...]

PR #691, opened 2024-12-09
Don't set a limit for image size

tsujan:
Qt6 started to set an allocation limit of 256 MiB. The behavior of Qt5 is restored here, i.e., no limit is set, because this is an image viewer.

Closes https://github.com/lxqt/lximage-qt/issues/690


tsujan:
Yes, it's about the same memory usage here > 256 MiB.

PR #692, opened 2024-12-09
Fixed invalid nullptr parameter in `QObject::connect`

tsujan:
`QNetworkAccessManager` is changed to a static `QPointer`, which is deleted when all instances of `Provider` are deleted.

In addition, all providers are created on the heap, as children of the upload dialog.

In this way, the following warning is silenced:

```
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter
```

### DIFF
--- a/src/mainwindow.h
+++ b/src/mainwindow.h
@@ -94,11 +94,11 @@ protected:
   virtual void resizeEvent(QResizeEvent *event);
   virtual void closeEvent(QCloseEvent *event);
 
+  virtual bool eventFilter(QObject* watched, QEvent* event);
+private Q_SLOTS:
   void onImageLoaded();
   void onImageSaved();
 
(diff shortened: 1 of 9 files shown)


tsujan:
@stefonarch 
Please first make sure you have nothing in `~/.config/QtProject/qtlogging.ini`, then run the app from a terminal and try to upload an image. You shouldn't see the above-mentioned warning in the terminal, and the uploading should be like before (that of imgur is broken because of recent changes in imgur).

stefonarch:
For some reason both upload options gives `.json` errors now, while ImgBB worked yesterday...

stefonarch (approved):

tsujan:
> works now, with IMG, no idea why. Error message is gone.

Frankly, I don't understand why an image viewer should have image uploading codes. Sites can change their rules or remove their services, and it's an ugly job for us to follow them.

PR #694, opened 2024-12-11
Fix off-by-one error in screenshot area selecting

isf63:
Because pixel coordinates start at zero the normal method for calculating rects is off by one.

Prior it was impossible to select pixels on the right and bottom edges.

### DIFF
--- a/src/screenshotselectareagraphicsview.cpp
+++ b/src/screenshotselectareagraphicsview.cpp
@@ -36,14 +36,15 @@ void ScreenshotSelectAreaGraphicsView::mousePressEvent(QMouseEvent *event)
     p0_ = QPointF(event->pos());
   } else {
     if(selectedAreaRect_ == nullptr) {
-      QColor highlight = palette().color(QPalette::Active,QPalette::Highlight);
+      QColor highlight = palette().color(QPalette::Active, QPalette::Highlight);
       QPen pen(highlight, 3, Qt::DashDotLine, Qt::RoundCap, Qt::RoundJoin);
       QColor color(highlight);
       color.setAlpha(128);
       QBrush brush(color);
       selectedAreaRect_ = scene()->addRect(QRectF(), pen, brush);
-    } 
-    selectedAreaRect_->setRect(QRectF(p0_,QPointF(event->pos())).normalized());
+    }
+    QPointF pos(event->pos().x() + 1, event->pos().y() + 1);
+    selectedAreaRect_->setRect(QRectF(p0_, pos).normalized());
   }
 }
 
@@ -54,5 +55,6 @@ void ScreenshotSelectAreaGraphicsView::mouseMoveEvent(QMouseEvent *event)
 
 void ScreenshotSelectAreaGraphicsView::mouseReleaseEvent(QMouseEvent *event)
 {
-  Q_EMIT selectedArea(QRectF(p0_,QPointF(event->pos())).normalized().toRect());
+  QPointF pos(event->pos().x() + 1, event->pos().y() + 1);
+  Q_EMIT selectedArea(QRectF(p0_, pos).normalized().toRect());
 }


tsujan:
> Because pixel coordinates start at zero…

That can't be the cause of what you see.

First please explain what you want to fix, and especially what you mean by "it was impossible to …". And please explain it in a way that I can understand it without logging into X11.

tsujan:
A rectangle whose top left and bottom right corners are at (0,0) and (2,2) respectively has a width and height of 3px, not 2 (see https://doc.qt.io/qt-6/qrect.html#bottom, for example).


Therefore, both rectangles of the code go beyond `event->pos()` by one pixel to the right and bottom *if `event->pos()` is really at the bottom right corner of the rectangle*; otherwise, see https://doc.qt.io/qt-6/qrect.html#QRect-1.

Is what you want to fix about the last scenario? If so, you need to check whether `event->pos()` is at the bottom right or not.

isf63:
I didn't take that case into consideration, I almost always made selections from top-left to bottom-right. Will fix.

isf63:
> please explain it in a way that I can understand it without logging into X11

It's currently not possible to select the pixel row and column of the far bottom and right. I noticed that when making a fullscreen selection and the image resolution was off by one from my monitor resolution.

tsujan:
> Will fix.

Please don't :)

Actually, it's a Qt6 bug. I'll explain it in an upcoming PR.

isf63:
FWIW this was my attempt:
```diff --git a/src/screenshotselectareagraphicsview.cpp b/src/screenshotselectareagraphicsview.cpp
index 7da45b7..9aaa656 100644
--- a/src/screenshotselectareagraphicsview.cpp
+++ b/src/screenshotselectareagraphicsview.cpp
@@ -43,8 +43,9 @@ void ScreenshotSelectAreaGraphicsView::mousePressEvent(QMouseEvent *event)
       QBrush brush(color);
       selectedAreaRect_ = scene()->addRect(QRectF(), pen, brush);
     }
-    QPointF pos(event->pos().x() + 1, event->pos().y() + 1);
-    selectedAreaRect_->setRect(QRectF(p0_, pos).normalized());
+    qreal pos_x = event->pos().x() + ((p0_.x() < event->pos().x()) ? 1 : -1);
+    qreal pos_y = event->pos().y() + ((p0_.y() < event->pos().y()) ? 1 : -1);
+    selectedAreaRect_->setRect(QRectF(p0_, QPointF(pos_x, pos_y)).normalized());
   }
 }
 
@@ -55,6 +56,7 @@ void ScreenshotSelectAreaGraphicsView::mouseMoveEvent(QMouseEvent *event)
 
 void ScreenshotSelectAreaGraphicsView::mouseReleaseEvent(QMouseEvent *event)
 {
-  QPointF pos(event->pos().x() + 1, event->pos().y() + 1);
-  Q_EMIT selectedArea(QRectF(p0_, pos).normalized().toRect());
+  qreal pos_x = event->pos().x() + ((p0_.x() < event->pos().x()) ? 1 : -1);
+  qreal pos_y = event->pos().y() + ((p0_.y() < event->pos().y()) ? 1 : -1);
+  Q_EMIT selectedArea(QRectF(p0_, QPointF(pos_x, pos_y)).normalized().toRect());
 }
```
 Seemingly worked but borders were a little off.

tsujan:
https://github.com/lxqt/lximage-qt/pull/695 should fix the problem. Your test would be very appreciated (I didn't log into X11).

I'm sorry that my previous comment was so hasty — I was busy with another code. Here is the whole story:

According to https://doc.qt.io/qt-6/qrect.html#normalized, `QRect::normalized()` should guarantee that the `QRect` includes its right and bottom edges. That was the case in Qt5. But Qt6 has a nasty regression in this regard, as the above-mentioned example code clearly shows.

Personally, I avoid some Qt methods in my codes because I don't trust them; `QRect::normalized()` is one of them, and this Qt6 bug confirmed my mistrust ;) It was why I thought that the same bug was present in lximage-qt. However, I was wrong.

Here we deal with `QRectF`, not `QRect`. Edge inclusion doesn't have meaning in the case of `QRectF`, but we need to be cautious when we want to make a `QRect` out of it.

PR #695, opened 2024-12-11
Fixed a small miscalculation in selection screenshot

tsujan:
Converstion of `QRectF` to `QRect` excludes the right and bottom edges. That's taken into account by the patch.

Also, the deprecated method `QMouseEvent::pos()` is replaced.

### DIFF
--- a/src/screenshotselectareagraphicsview.cpp
+++ b/src/screenshotselectareagraphicsview.cpp
@@ -33,7 +33,7 @@ ScreenshotSelectAreaGraphicsView::ScreenshotSelectAreaGraphicsView(QGraphicsScen
 void ScreenshotSelectAreaGraphicsView::mousePressEvent(QMouseEvent *event)
 {
   if(p0_.x() < 0) {
-    p0_ = QPointF(event->pos());
+    p0_ = event->position();
   } else {
     if(selectedAreaRect_ == nullptr) {
       QColor highlight = palette().color(QPalette::Active,QPalette::Highlight);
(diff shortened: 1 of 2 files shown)


isf63:
Works as intended, able to select the right and bottom.

src/screenshotselectareagraphicsview.cpp
```diff
@@ -42,8 +42,8 @@ void ScreenshotSelectAreaGraphicsView::mousePressEvent(QMouseEvent *event)
       color.setAlpha(128);
       QBrush brush(color);
       selectedAreaRect_ = scene()->addRect(QRectF(), pen, brush);
-    } 
-    selectedAreaRect_->setRect(QRectF(p0_,QPointF(event->pos())).normalized());
+    }
+    selectedAreaRect_->setRect(QRectF(p0_,event->position()).normalized());
```
isf63, line 46: The painted preview rect should also have margins. You can see the difference in position by comparing the drawn edges at the far bottom-right with the top-left.

tsujan: If by "should" you mean it has a *visible* margin, that may be a result of rounding because It's a `QRectF`. But if you mean we should add a margin here, no, we shouldn't, again because it's a `QRectF`.

Anyway, if there's also another miscalculation, it should be found and fixed in another PR.

tsujan: Also, see https://doc.qt.io/qt-6/qgraphicsrectitem.html#details. That's exactly what the code does. Qt is responsible for a probable inconsistency. Since we don't know when it might be corrected, we can't add margins. Fortunately, it isn't important for us in practice either.

This PR was about *our* miscalculation.

PR #697, opened 2025-01-01
Ensure current thumbnail is in visible part of thumbnail bar

tsujan:
Previously, the thumbnail of the current image might be outside the visible part of the thumbnail bar when the app was started, although the code tried to scroll to it.


stefonarch:
I can't reproduce the issue.

tsujan:
> I can't reproduce the issue.

Yes, expected, because it's random and rare.

Anyway, `QCoreApplication::processEvents();` should have been removed from the code, as in this patch.

Issue #698, opened 2025-01-28
Problem in rectangular annotation


tsujan:
The rectangular annotation won't be shown if

 1. The image is shown with its original size, and
 2. The rectangle is drawn from right to left or bottom to top.

But it'll be shown if the image is zoomed out or in.

PR #699, opened 2025-01-28
Fixed a problem in rectangular annotation

tsujan:
Fixes https://github.com/lxqt/lximage-qt/issues/698


stefonarch (approved):
Issue is fixed, annotations all working now.

Issue #701, opened 2025-01-29
"original size" incorrect with scaled monitors (Wayland)


stefonarch:
This happens only when one monitor is scaled up, like 110% or 125% but can be seen on both monitors identical.

##### Expected Behavior

On a screen 1920x1080 px an image which is 1920px width should fill it.
##### Current Behavior

It'about half the screen,
Qruler has the same issue, other Qt apps (phototonic, photoQ, falkon) are fine. X11 is fine.

##### Steps to Reproduce (for bugs)

1. Take a full screenshot under Wayland
2. Open the result with LXimage-qt

##### System Information

* LXimage-qt version: git
* Distribution & Version: arch, latest
* Qt Version: 6.8.1-1

isf63:
I get the same behavior on X11 and Wayland with a fullscreen screenshot, but the window is about 4/5 the width of my monitor. If LXImage-Qt is maximized and closed, it maximizes the window upon reopening, otherwise it tries to pick to a size.

tsujan:
If it's half the screen, your WM has a problem. The window is shown such that it fits the available area, giving enough space to some title-bars/decorations that may be thick, like this:

![Image](https://github.com/user-attachments/assets/d59ab142-0405-4427-aaa5-3b254e1b730d)

tsujan:
Please close this. This behavior is the best that can be achieved.

stefonarch:
This is kwin_wayland, monitor 1 scaled to 125%, monitor 2 100%  and this is a fullscreen window with "Original size" on monitor 2. 

![Image](https://github.com/user-attachments/assets/09690c15-b882-4e4d-b3ca-f00cce1b49c6)

tsujan:
You cheated; this is a totally different report.

It may be about the same Qt+Wayland issue that I encountered and circumvented again and again. Will look into it. Thanks.

stefonarch:
I didn't cheat - I didn't have the full picture, and didn't think 110% would make 200% ;)

tsujan:
> didn't think 110% would make 200% on both monitors ;)

Smells like that old `QScreen` issue on Wayland.

tsujan:
I don't have my external monitor, and there's an information shortage here. The patch I gave to you privately should work.

 1. Did you start lximage-qt on the first monitor?
 2. Is there another lximage-qt window open on the first monitor?
 3. Did you move the window from the first monitor to the second?
 4. What happens on the first monitor?

[...]

PR #702, opened 2025-01-30
Use the device pixel ratio of the window on Wayland

tsujan:
On Wayland, each screen can have its own device pixel ratio, which is correctly reported by the window, not by the app.

The app's behavior under X11 is like before.

Closes https://github.com/lxqt/lximage-qt/issues/701

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -32,6 +32,7 @@
 #include <QPainter>
 #include <QPainterPath>
 #include <QGuiApplication>
+#include <QWindow>
 #include <QtMath>
 
 #define CURSOR_HIDE_DELY 3000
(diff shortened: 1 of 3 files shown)


stefonarch (commented):
This looks ok, but when the window is moved to another monitor with a different scaling artefacts can be seen resizing and adapting, until the image is reloaded (rotate looks like reload).

https://github.com/user-attachments/assets/af8eee5f-01e5-4a45-81d0-c0e37700aeb7

tsujan:
With all compositors? I ask to rule out compositor-specific issues.

stefonarch:
What is real weird to me: working with full screenshots. When I open them from pcmanfm-qt I get the above behavior and "Original size" is again half (see the white background in the video), but when I open them from lximage-qt by "Recent files" behavior is ok. Also rotating them fixes it.


https://github.com/user-attachments/assets/2a89b13f-bcf4-4a82-bc1e-6c9bb83e09c4

tsujan:
> What is real weird to me:…

Actually, this one isn't weird to me. I'm not sure if I should try to fix it here or in another PR, because it's another issue. What do you think?

stefonarch:
> With all compositors? I ask to rule out compositor-specific issues.

Good catch. Tested with kwin_wayland, no issue there.

stefonarch (approved):

tsujan:
> Tested with kwin_wayland, no issue there.

I have a theory that I need to test, but only with your help. So, I suggest we merge this PR first — of course, only if you confirm that it does its job.

Issue #703, opened 2025-01-30
Unusual behavior when moving window to another screen under Wayland


tsujan:
Problems happen when the other screen has a different scaling. All problems go away as soon as the image is reloaded, but read @stefonarch's descrition: https://github.com/lxqt/lximage-qt/pull/702#pullrequestreview-2584881181

PR #704, opened 2025-01-31
Fixes and workarounds for multi-screen scaling issues on Wayland

tsujan:
The issues were bizarre — especially in one case, all relevant Qt methods gave wrong pixel ratios — but workarounds were effective.

Also,
 * Fixed an old issue in showing GIF animations with pixel ratios > 1;
 * Commented out "RememberWindowSize" from settings because it was never used (but might be used later); and
 * Removed a redundant private variable.

Closes https://github.com/lxqt/lximage-qt/issues/703

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -51,7 +51,7 @@ ImageView::ImageView(QWidget* parent):
   scaleFactor_(1.0),
   autoZoomFit_(false),
   smoothOnZoom_(true),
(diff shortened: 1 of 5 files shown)


stefonarch (approved):

tsujan:
Merging quickly, before you change your mind ;)

The workarounds used here should work with any Qt image viewer — until Qt becomes better at handling Wayland.

PR #714, opened 2025-04-19
Update root translation: "&Recently Opened Files" → "&Recent Files"

isf63:


tsujan (approved):
This is shorter and better. Thanks.

GTM.

PR #720, opened 2025-05-18
Cleanup `-d` screenshot filename generation

isf63:
```
  -d, --dirscreenshot <DIR>  Take a screenshot and save it to the directory
                             without showing the GUI (deprecated, please use
                             screengrab instead)
```

Previously the filenames were incomprehensible without knowledge of the code.

This also simplifies things.

### DIFF
--- a/src/screenshotdialog.cpp
+++ b/src/screenshotdialog.cpp
@@ -229,14 +229,6 @@ void ScreenshotDialog::doScreenshot() {
   deleteLater(); // destroy ourself
 }
 
-static QString buildNumericFnPart() {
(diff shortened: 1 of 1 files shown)


tsujan:
> Previously the filenames were incomprehensible

They weren't supposed to. The goal was explained in the code clearly.

I don't see a reason for a change.

isf63:
If anything this PR is about simplification - the filename is sane, and the code is cleaner/simpler

tsujan:
> … is about simplification

Too subjective. We have a clean static function that does the job for us and includes the pid too.

Issue #721, opened 2025-05-19
[Feature request] Multipage tiff navigation toolbar/toolset


sfzaw:
**Is your feature request related to a problem? Please describe.**
Lximage-qt is unable to view Multipage tiff image properly. To be exact it only shows the first page. below contains example of such file. (will provide personal examples if asked)
https://www.nightprogrammer.org/development/multipage-tiff-example-download-test-image-file/

**Describe the solution you'd like**
A permanent/pop up navigation toolbar that shows page number and left+right ( <-- [6/10] --> ) toolbar 

**Describe eventual alternatives you've considered**
hardly any linux image viewer support this. so windows apps on wine is good alternative

#### Context
Large archival projects, Scanned Documents, contracts, certificates etc. For example, i am digitizing 70 year old land documents on a Toshiba AIO photocopier. And each set provides output with multi-page tiff formats. If i choose to do it single page that means setting up the machine manually for every single page. Choosing PDF, Jpeg, png will amount to data loss which halves the purpose of digitizing and archival. I am willing to provide such images if needed.

stefonarch:
Gimp opens the small test file with pages as levels, I see. Not very handy but the pages can be seen.

tsujan:
Sorry, out of scope for us.

stefonarch:
ImageMagick `display multipage_tiff_example.tif` press space button:

https://github.com/user-attachments/assets/b3292bb7-66be-4ac9-ae21-c7062bf3685d

tsujan:
> ImageMagick display multipage_tiff_example.tif press space button:

Nice.

An advanced viewer like Gwenview doesn't show them either. The fact it that lximage-qt is based on what Qt supports, plus the file management provided by GLib . We don't to go beyond that here.

sfzaw:
but qview can show them through 1 sec slide show has same dependencies as lximage-qt. if space can toggle the multipage thats good enough for now....

tsujan:
> but qview can show them through 1 sec slide show(too fast; doesnt stop) has same dependencies as lximage-qt.

Then reopened, for any contributor who might want to implement it without extra dependencies — as the maintainer of this app, I don't have the motivation or time for now.

tsujan:
As a side note, your report made me see a flaw in our kwin_wayland backend of lxqt-panel and fix it — when I opened one of those images in GIMP :)

tsujan:
@sfzaw
[...]

PR #723, opened 2025-05-26
Supported multi-page TIFF images

tsujan:
If a multi-page TIFF image is opened, the menu-items "Next Frame" and "Previous Frame" will be added to the "Go" menu, with the shortcuts "Ctrl+Right" and "Ctrl+Left" respectively. Of course the shortcuts can be changed in the shortcut editor.

To prevent cluttering of the GUI for this special use case, I didn't add another way of informing the users. They should either check the "Go" menu or press those shortcuts to see whether a TIFF image includes multiple pages.

Closes https://github.com/lxqt/lximage-qt/issues/721

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -506,7 +506,26 @@ void ImageView::setImage(const QImage& image, bool show, bool updatePixelRatio) 
   queueGenerateCache();
 }
 
-void ImageView::setGifAnimation(const QString& fileName) {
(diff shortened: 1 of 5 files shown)


tsujan:
> Does is work also for animated gifs?

I made sure it doesn't ;) Because animated GIF images should be animated.

tsujan:
> we should document it

Yes, concisely: "A minimal support for …".

Issue #727, opened 2025-05-27
Annotations aren't copied to clipboard


tsujan:
Very easy to fix. I'll do it after https://github.com/lxqt/lximage-qt/pull/726 is decided.

Issue #192, opened 2018-11-03
"Non-Solid" Background


tsujan:
##### Expected Behavior
An option for removing the solid background and showing the image like this (not an lximage-qt screenshot):

![solid](https://user-images.githubusercontent.com/981076/47955439-6284b280-dfad-11e8-9349-8a144f8f21db.png)

It is useful especially for checking the translucent parts of the image (like shadows).

##### Current Behavior
Currently, there's only a solid background.

##### Possible Solution
I've done it elsewhere; should look (again) into lximage-qt's code.

PR #728, opened 2025-05-27
Fixed copying annotations to clipboard

tsujan:
Actually, a redundant private variable, that could become a source of trouble, has been removed, and as a result, copying annotations is fixed automatically.

Fixes https://github.com/lxqt/lximage-qt/issues/727

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -61,7 +61,6 @@ MainWindow::MainWindow():
   QMainWindow(),
   contextMenu_(new QMenu(this)),
   slideShowTimer_(nullptr),
-  image_(),
   // currentFileInfo_(nullptr),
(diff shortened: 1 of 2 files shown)


stefonarch (approved):
Couldn't find any problem, I'm disappointed and sorry ;)
Only ctrl+z doesn't remove the last added anotation(s) which would be a nice feature too, instead of reloading and redoing,

tsujan:
> Only ctrl+z doesn't remove the last added anotation(s)

Annotations history isn't implemented. It needs a lot of extra codes for a job that belongs to image editors.

> I'm disappointed and sorry ;)

Sorry that I'm happy ;) Implementing https://github.com/lxqt/lximage-qt/issues/192 after this…

PR #729, opened 2025-05-28
Support non-solid (tiled) background

tsujan:
It's especially useful for checking the translucent parts of images that support translucency.

The user can set it temporarily by toggling its action in "View" → "Solid Background" (its shortcut can be defined in the shortcut editor) or change the default by toggling its new option in Preferences → Image.

Closes https://github.com/lxqt/lximage-qt/issues/192

### DIFF
--- a/src/imageview.cpp
+++ b/src/imageview.cpp
@@ -54,7 +54,8 @@ ImageView::ImageView(QWidget* parent):
   isSVG_(false),
   currentTool_(ToolNone),
   nextNumber_(1),
-  showOutline_(false) {
+  showOutline_(false),
+  tiledBg_(false) {
 
   setViewportMargins(0, 0, 0, 0);
   setContentsMargins(0, 0, 0, 0);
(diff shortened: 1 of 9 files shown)


stefonarch:
Just wanted to merge updated *.ts files but this will add more as it seems.

tsujan:
> but this will add more as it seems

Yes. I think no more translatable strings will be added for a while after it.

isf63:
In English AFAIK "frame" is correct. There are "Frame"/"Page"/"Slide", of which "frame" is most relevant to multiple sub-images in a computer file.

stefonarch:
immagine

tsujan:
It's like saying that we should use "Register card" instead of "Tab" because it's "Registerkarte" in German ;) There are so many examples.

tsujan:
> I won't give more battle for this

Good, because if you did, I'd have to say nonsense like "Page is used only for PDF and shouldn't be used for images," just to win it ;)

stefonarch:
I stopped only because I know that you can't lose battles usually....

PR #744, opened 2025-07-31
Fix showing of ampersand in recent files menu

isf63:
Exactly the same as https://github.com/lxqt/pcmanfm-qt/pull/2081 but for LXImage-Qt


isf63:
Can confirm there are glitches with opening files with ampersands from the Recent Files menu. Both with display names and failing to load.

PR #758, opened 2025-09-27
Set window icon

City-busz:
This ensures that the icon is set on all windows on X11.

Similar to https://github.com/lxqt/pcmanfm-qt/pull/2096, the non-modal preferences window has no icon on X11 without this.

### DIFF
--- a/src/application.cpp
+++ b/src/application.cpp
@@ -37,6 +37,7 @@ Application::Application(int& argc, char** argv):
   libFm(),
   windowCount_(0) {
   setApplicationVersion(QStringLiteral(LXIMAGE_VERSION));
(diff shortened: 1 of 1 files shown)


tsujan (approved):
It should be safe.

isf63:
While Openbox is technically dead, unfortunately some distros (Arch Linux being the main one, Manjaro too?) include it in the LXQt metapackage. It also is valid for very weak computers.

tsujan:
LXQt has no problem with any WM. If Openbox still works relatively fine, why not using it? But changing something just because this or that WM has a problem isn't an option. The current PR and its pcmanfm-qt counterpart were more general than that.

PR #760, opened 2025-09-30
Assign the print dialog to the parent window

City-busz:
This ensures that the dialog is attached to the main window.


tsujan:
> I see some that lack a parent argument 

Not all dialogs of an app should have parent widgets. Please don't go to extremes ;)

PR #788, opened 2026-05-13
feat: add open in new window

Chiitoo:
Wonder if something like this would be to user-specific to add in?

There's currently also some code to add a button but that can be disregarded.

To put it in words, this adds an option to open the current image in a new window, both in the right-click context menu, as well as via a shortcut.

As a happy accident, the way I did it makes the image open in the exact size the last image was, which is a bit weird but kind of convenient...

### DIFF
--- a/src/mainwindow.cpp
+++ b/src/mainwindow.cpp
@@ -187,6 +187,8 @@ MainWindow::MainWindow():
   contextMenu_->addAction(ui.actionRotateCounterclockwise);
   contextMenu_->addAction(ui.actionFlipHorizontal);
   contextMenu_->addAction(ui.actionFlipVertical);
+  contextMenu_->addSeparator();
+  contextMenu_->addAction(ui.actionOpenInNewWindow);
 
   auto sortGroup = new QActionGroup(ui.menu_View);
   sortGroup->setExclusive(true);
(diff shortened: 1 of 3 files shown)


tsujan:
Thanks, but I'm not sure this is needed — frankly, it seems redundant and cluttering to me. The user can use the thumbnail view and open any image in a new window or even with another app. In addition, we already have "New Window"; and there are more ways … to say nothing of the fact that opening an image in a new window isn't something most users need to do frequently.

The code needs changes too, but that's another subject.

Chiitoo:
Yeah, I did have the feeling that it's a bit too specific for my workflow, which indeed might not be common at all.

Figured I'd still offer it and see what others think.  :]

Previously I'd use `File -> Open With... -> LXImage-Qt` but it is a bit cumbersome when doing it a lot.

Using thumbnails makes that part better, but doesn't feel quite as nice as being able to use a shortcut, the context menu, or a separate button.

As for `New Window`, that only opens a new empty window, so not quite what I need.

In any case, thanks!

tsujan:
> As for New Window, that only opens a new empty window, so not quite what I need.

Examples:

 Open a new window and drag-and-drop from the thumbnail bar of the first window.
 Open a new window and copy/paste from clipboard.

PR #790, opened 2026-05-24
Add AVIF to supported mime types

cunlem:


isf63:
AVIF and HEIF require Qt supplementary libraries to work. I don't know the details here, maybe @tsujan knows if default support should be added.

tsujan:
Thanks but we don't add every possible mimetype that may be supported by qt6-imageformats, kimageformats or other plugins. We leave them to the user instead (→ File Associations in LXQt, similar tools elsewhere).

cunlem:
Aren't some of listed (tga, pcx) already depending on kimageformats?

tsujan:
If so, they were added only because many users insisted. This should be stopped somewhere because the number of *potentially* supported mimetypes is not only great but increasing. I also remember that a user even wanted to make it depending on the types supported at the compilation time, which was out of the question.

cunlem:
Ok I feel like having potential mime types listed that turn up unsupported at runtime and simply displaying error is not bad UX. I prefer this to launching GIMP. The image formats will become more widespread since they're supported by web browsers now.

tsujan:
> I feel like having potential mime types listed that turn up unsupported at runtime and simply displaying error is not bad UX.

It would cause more headache, both for users and for us ;) Users might get confused and report "bugs". In addition, we can't know beforehand which distro includes which (version of) plugin in its repository. There are cases like PDF too, and more.

Don't get me wrong; your good intention is appreciated.

